PR #6, opened 2018-07-25
Change commented-out config to explicit False.

johannct:
Explicitly set config values instead of relying on defaults.
Add notes that we intend to turn these on in the future.


johannct:
Merging this straightforward proposal.

PR #9, opened 2018-08-27
Add focalplane summary script.

SimonKrughoff:
@RobertLuptonTheGood do you mind having a look at this?  If this can merge, we can get rid of the DESC fork of `obs_lsstSim` all together.


kadrlica:
I've tried to run this on some DC2 data that I ingested with `obs_lsstCam` and I get this error:

kadrlica:
I also think that there is some issue with the zscale being set in the case that some sensors have no data (see below).

image 3

python/lsst/obs/lsstCam/simbutlerimage.py
```diff
@@ -0,0 +1,62 @@
...
+
+        im = None
+        cid = ccd.getName()
+        did = parse_name_to_dataId(cid)
+        if self.butler is not None and did['raftName'] not in ['R00', 'R04', 'R40', 'R44']:
+            self.kwargs.update(did)
+            try:
+                im = self.butler.get(self.type, **self.kwargs)
+                ccd = im.getDetector()  # possibly modified by assembleCcdTask
+                if self.type == 'eimage' or self.type == 'calexp':
```
kadrlica: Do we want to be this restrictive?

python/lsst/obs/lsstCam/simbutlerimage.py
```diff
@@ -0,0 +1,62 @@
...
+        if im is None:
+            if not as_masked_image:
+                im = self._prepareImage(ccd, imageFactory(*bbox.getDimensions()), binSize)
+                return rotateImageBy90(im, ccd.getOrientation().getNQuarter()), ccd
+            else:
+                im = self._prepareImage(ccd, afwImage.makeMaskedImage(imageFactory(*bbox.getDimensions())),
+                                          binSize)
+                return rotateImageBy90(im, ccd.getOrientation().getNQuarter()), ccd
+
+        if self.type == "raw":
```
kadrlica: Didn't we enforce above that `self.type` couldn't be "raw"?

kadrlica:
I've been having trouble getting `makeFpSummary.py` (the `obs_lsstSim` version) run with wildcard `--id` arguments. The following works with `--show data`, but when I attempt to run for real it exits after the first visit. I'm not sure if this is a problem isolated to `obs_lsstSim`, in which case it can be ignored.

```
makeFpSummary.py $OUTDIR --dstype eimage --id visit=181000..186000 --output $OUTDIR/output
```

bin.src/makeFpSummary.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def runDataRef(self, expRefList, butler):
+        """Make summary plots of full focalplane images.
+        """
+        sbi = SimButlerImage(butler, type=expRefList[0].butlerSubset.datasetType, visit=expRefList[0].dataId['visit'])
+
+        for expRef in expRefList:
+            data_id = expRef.dataId
+            ccd = butler.get('calexp_detector', **data_id)
+            try:
```
kadrlica: Should this rotation be different for `eimage` vs `calexp`? @RobertLuptonTheGood suggested on Slack that the orientation is different.

RobertLuptonTheGood:
Can we not just fix the eimages once and for all in a pre-burner (assuming that we can't fix phosim)?  It's just confusing everyone, and making code more complex, by not making them look like "real" data.  If we did that, all the tooling would Just Work (TM).

SimonKrughoff:
@RobertLuptonTheGood I think we should just close so we can focus on #20. What do you think?

SimonKrughoff:
Closing this in preference of other PRs.

PR #11, opened 2018-08-28
add Robert's suggested workaround for the median-bug in the ISR overscan correction

jchiang87:
The need for these changes was noted by Robert in the [desc-dm-dc2 channel](https://lsstc.slack.com/archives/C978LTJGN/p1535461457000100)


heather999:
@rearmstr I've assigned you as a reviewer on this PR, we need these updates for processing the Run1.2i data. If you haven't seen the discussion on the desc-dm-dc2 channel, this might be a good starting point: https://lsstc.slack.com/archives/C978LTJGN/p1535461457000100

PR #15, opened 2018-09-04
U/krughoff/make fp fixups

heather999:
Preparing to pull in the makeFpSummary updates for Run1.2


kadrlica:
This was being tracked in #9, but Robert suggested that we could change the eimage orientation so that it matches the calexps. Right now it's pretty confusing that the FP images have different layouts.

heather999:
superseded by #20

PR #17, opened 2018-09-05
DM-6640: IsrTask is not a valid CmdLineTask

czwa:
This ticket adds a stand-alone runIsr.py script, as well as ensuring that runIsr.py and processCcd.py read the same ISR configuration files.

### DIFF
--- a/config/processCcd.py
+++ b/config/processCcd.py
@@ -1,13 +1,17 @@
+"""
+LSST Cam-specific overrides for ProcessCcdTask
+"""
 import os.path
+
 from lsst.utils import getPackageDir
 from lsst.meas.algorithms import LoadIndexedReferenceObjectsTask
 from lsst.meas.algorithms import ColorLimit
 
-configDir = os.path.join(getPackageDir("obs_lsstCam"), "config")
+obsConfigDir = os.path.join(getPackageDir("obs_lsstCam"), "config")
 
-bgFile = os.path.join(configDir, "background.py")
+config.isr.load(os.path.join(obsConfigDir, "isr.py"))
 
-config.load(os.path.join(getPackageDir("obs_lsstCam"), "config", "lsstCam.py"))
+bgFile = os.path.join(obsConfigDir, "background.py")
 
 # Cosmic rays
 config.charImage.repair.cosmicray.nCrPixelMax = 1000000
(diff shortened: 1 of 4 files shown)


config/processCcd.py
```diff
@@ -1,13 +1,17 @@
...
+LSST Cam-specific overrides for ProcessCcdTask
+"""
 import os.path
+
 from lsst.utils import getPackageDir
 from lsst.meas.algorithms import LoadIndexedReferenceObjectsTask
 from lsst.meas.algorithms import ColorLimit
 
-configDir = os.path.join(getPackageDir("obs_lsstCam"), "config")
+ObsConfigDir = os.path.join(getPackageDir("obs_lsstCam"), "config")
```
yalsayyad: https://developer.lsst.io/python/style.html#naming-conventions-for-science-pipelines
What was wrong with `configDir`?

czwa: Naming convention fixed.
After going through the various `obs_*` packages, and seeing each one using a different name for this same concept, I unified them to a single name for my own sanity if nothing else.  @PaulPrice was also not a fan of this change to `obsConfigDir`.  Is there a consensus that these should be left different in each package?

mfisherlevine: I like the unification of the naming across packages, but have no strong feelings about whether to use `obsConfigDir` or `configDir`. I'd lean slightly towards the latter, but whatever is the least work is also fine 😉

mfisherlevine:
General comment: `processCcd.py` in this package is a bit of a mess - an `import` in the middle of the config, no consistency with spaces are `=` operators, etc. You didn't do this, but when I looked in the file it made me sad.

heather999:
So how do we get processCcd.py cleaned up @mfisherlevine?   Can a mere mortal do that? :)

mfisherlevine:
@heather999 Yes, just file a ticket and it'll get done, it should be trivial to fix the things I mentioned here, but here is not the place to have a conversation about that 😉

mfisherlevine:
@heather999 As you can [see here](https://github.com/lsst/obs_lsstCam/pull/17#pullrequestreview-154182392) I did ask for this change but it looks like it was forgotten, so that's probably going to need removing somewhere else as this has now merged.

PR #18, opened 2018-09-06
add crosstalk values from Adam's measurements

jchiang87:
We'd like to use these coefficients for Run2.0i.


heather999:
@jchiang87 is this still relevant and do we need to get this into master?

jchiang87:
these are still relevant and we need to get them into master.

PR #20, opened 2018-09-11
U/rhl/make fp fixups

RobertLuptonTheGood:
Misc changes starting from `u/krughoff/makeFp_fixups`, which this PR supersedes by agreement with Simon.


python/lsst/obs/lsstCam/lsstCamMapper.py
```diff
@@ -181,7 +182,7 @@ def assemble_raw(dataId, componentInfo, cls):
 
     try:
         exposure = _lsstCamMapper.std_raw(exposure, dataId)
-    except:
+    except:  # noqa E722
```
SimonKrughoff: Is there a reason for the bare except?  Surely it should at least be `except Exception`, right?

timj: You should never mask this. It needs to be fixed to decide between `Exception` and `BaseException`.

RobertLuptonTheGood: I don't understand this.  I want to catch any problem, so I can change this to catch `(Exception, BaseException)` if you like, but I think that the `noqa E722` is clearer.

timj: Python 3 changed everything. Bare except is now deprecated because it's not entirely clear what you mean. If you mean *everything* then you say you mean everything by using `BaseException`. If you don't mean catching Ctrl-C and `sys.exit` then you want to use `Exception` instead.

.gitignore
```diff
@@ -6,3 +6,4 @@ config.log
 bin
 ups/*.cfgc
 python/lsst/obs/lsstCam/version.py
+bin
```
SimonKrughoff: `bin` is already in this list.  Did you mean to add it twice?

PR #30, opened 2018-10-16
DM-16215: Change instFlux->gaussianFlux following DM-10302

hsinfang:


config/forcedPhotCcd.py
```diff
@@ -6,4 +6,4 @@
 config.measurement.load(os.path.join(getPackageDir("obs_lsst"), "config", "kron.py"))
 config.load(os.path.join(getPackageDir("obs_lsst"), "config", "cmodel.py"))
 
-config.measurement.slots.instFlux = None
+config.measurement.slots.gaussiaFlux = None
```
parejkoj: `gaussianFlux`

hsinfang: Thanks for catching my typo! Apparently I didn't test it...

parejkoj: Are there any tests in this package?

hsinfang: None as I know of. But I could test them manually in principle...

hsinfang: There are some preliminary ones on disk. I haven't tried them out though.

parejkoj: Why are there no tests in this package? Every other obs package has them (including obs_lsstSim, which this is ostensibly replacing).

hsinfang: I see this as work in progress, and not officially ready for its prime time yet.  It's not yet included in `lsst_distrib`, and it will be renamed.

parejkoj: But it should have tests from the very start!

Anyway, not a problem for this ticket, but it is a serious issue with the package.

PR #31, opened 2018-10-18
DM-16236 -- Make it possible to run corner chips through ISR

SimonKrughoff:


python/lsst/obs/lsst/phosim.py
```diff
@@ -73,11 +73,7 @@ def translate_filter(self, md):
         filter : `str`
             filter name
         """
-        raftName = self.translate_raftName(md)
-        if raftName in ["R00", "R04", "R40", "R44"]:
-            return "NONE"
-        else:
-            return md.get("FILTER")
+        return md.get("FILTER")
```
mfisherlevine: The code change itself seems fine, but I'm kind of curious about why these were deliberately and specifically excluded before. Do you have any idea why that is, and hence why this change is necessary? I just worry that it it's far from an accident that's being reversed here.

policy/lsstCamMapper.yaml
```diff
@@ -203,6 +203,18 @@ datasets:
     python: lsst.pipe.tasks.indexReferenceTask.IngestIndexedReferenceConfig
     storage: ConfigStorage
     template: config/IngestIndexedReferenceTask.py
+  runIsr_config:
```
mfisherlevine: Perhaps none of these were in `obs_base` when you made these additions, but they seem to be now, so I think this can probably be cut down to only specifying the template for the `runIsr_metadata`, the rest might just work now? Otherwise, see the individual comments on the rest of this block.

policy/lsstCamMapper.yaml
```diff
@@ -203,6 +203,18 @@ datasets:
...
+    persistable: Config
+    python: lsst.meas.mosaic.isrTask.IsrTaskConfig
```
mfisherlevine: Why does this differ from the nominal `lsst.ip.isr.RunIsrConfig` specified in `obs_base`?

policy/lsstCamMapper.yaml
```diff
@@ -203,6 +203,18 @@ datasets:
...
+    storage: ConfigStorage
+    tables: raw
```
mfisherlevine: Does this need a table? One isn't specified in `obs_base`...

policy/lsstCamMapper.yaml
```diff
@@ -203,6 +203,18 @@ datasets:
...
+    template: config/isr.py
```
mfisherlevine: Shouldn't this by `runIsr.py`?

policy/lsstCamMapper.yaml
```diff
@@ -203,6 +203,18 @@ datasets:
...
+  runIsr_metadata:
+    persistable: PropertySet
+    python: lsst.daf.base.PropertySet
+    storage: YamlStorage
+    tables: raw
+    template: metadata/Isr.yaml
```
mfisherlevine: Should this be a capital `I` in `Isr.yaml`?

policy/lsstCamMapper.yaml
SimonKrughoff: OK.  This is right.  I've reverted the change, so now it only sets the metadata.  Hopefully this helps.

PR #32, opened 2018-10-23
Added flake8 config and fixed flake8 warnings

womullan:


timj:
```
./python/lsst/obs/lsst/ingest.py:179:16: W605 invalid escape sequence '\S'
```
probably needs fixing as well for the new flake8.

timj:
Please also add a `.travis.yml` file so that when we turn on travis things will work properly:
```yaml
sudo: false
language: python
matrix:
  include:
    - python: '3.6'
      install:
      - pip install flake8
      script: flake8
```

PR #33, opened 2018-10-24
Requesting a merge of DC2 Run1.2x configs

heather999:
@johannct @boutigny @rearmstr @hsinfang We are trying to merge some remaining configuration parameters that we have been using for Run1.2x processing, before this package is officially part of the DMstack.  One change @jchiang87  I'm not sure we should merge to master is the workaround for the median-bug, as this might now be fixed in the later versions of the stack? https://github.com/lsst/obs_lsstCam/commit/5a17448f1c7db41246aee8a55af61ce4edd108dd
Is there anything missing that we should be sure to include that may be in use currently in the Run1.2p processing, Johann?

### DIFF
--- a/config/isr.py
+++ b/config/isr.py
@@ -6,5 +6,10 @@ if hasattr(config, 'ccdKeys'):
 
 config.doLinearize = False
 config.doDefect = False
-config.doCrosstalk=True
+# skip xtalk for Run1.2x
+config.doCrosstalk = False
 config.doAddDistortionModel = False
+
+# Work-around for median-bug in overscan correction (DM-15203).
+config.overscanFitType = 'POLY'
+config.overscanOrder = 0
--- a/config/coaddDriver_noPSF.py
+++ b/config/coaddDriver_noPSF.py
@@ -1,0 +1,2 @@
+from lsst.pipe.tasks.selectImages import WcsSelectImagesTask
+config.select.retarget(WcsSelectImagesTask)
--- a/config/forcedPhotCcd.py
+++ b/config/forcedPhotCcd.py
@@ -6,4 +6,4 @@ config.measurement.load(os.path.join(getPackageDir("obs_lsst"), "config", "apert
 config.measurement.load(os.path.join(getPackageDir("obs_lsst"), "config", "kron.py"))
 config.load(os.path.join(getPackageDir("obs_lsst"), "config", "cmodel.py"))
 
-config.measurement.slots.instFlux = None
+config.measurement.slots.gaussianFlux = None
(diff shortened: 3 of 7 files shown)


johannct:
I have been running with vanilla branch `dc2/run1.2i`. So if everything has been ported from one branch to the other, I am fine. As reported by @rearmstr in #dm_lsstcam we need to take a decision about what to change in lsstCamMapper.yaml to get rid of the "-"/"_"" discrepancy. I agree with @TallJimbo that we should not override defaults if we do not have to. 

Another thing is the makeSkyMap.py config :

heather999:
I purposely pulled in the updates from dc2/run1.2i into this branch - but I'll double check @johannct 
Agreed we need to fix lsstCamMapper.yaml - is it enough remove all the entries that refer deepCoadd_results in lsstCamMapper.yaml?  It looks to me that those entries came from Ian as part of our updates to bring this package in line with problems we previously encountered with the processing: https://github.com/lsst/obs_lsstCam/pull/12/files

config/isr.py
```diff
@@ -6,5 +6,10 @@
 
 config.doLinearize = False
 config.doDefect = False
-config.doCrosstalk=True
+# skip xtalk for Run1.2x
+config.doCrosstalk = False
 config.doAddDistortionModel = False
+
+# Work-around for median-bug in overscan correction (DM-15203).
+config.overscanFitType = 'POLY'
```
jchiang87: As noted by Heather in the comments on this PR, the changes that these lines are meant as a work around are already in the weeklies so overscan stuff here can be omitted.

config/isr.py
```diff
@@ -6,5 +6,10 @@
 
 config.doLinearize = False
 config.doDefect = False
-config.doCrosstalk=True
+# skip xtalk for Run1.2x
+config.doCrosstalk = False
```
jchiang87: We only want to disable crosstalk for Run[1.2,2.0]i data, and that's already handled for imsim in https://github.com/lsst/obs_lsstCam/blob/dc2/run1.2-v2/config/imsim/processCcd.py, so this change should be omitted.

config/isr.py
heather999: Just want to make sure I remove the right bit.. this comment doesn't appear to include the config.overscanOver=0 line..  I would have been inclined to remove that too. Or was that unrelated to the median bug @jchiang87?

jchiang87: Yes, it includes both lines as they were both in the original PR I made.  In fact, all of the changes to this file in this PR can be omitted.

heather999: Hope I fixed this correctly :)

heather999:
I've rebased and I believe I've addressed @jchiang87's comments and removed the median-bug in overscan correction and no changes have been made to the doCrosstalk setting.  Dominique did inquire about this last bit though: https://lsstc.slack.com/archives/C978LTJGN/p1542302679184600  On master "doCrosstalk=True" is the default and then we would have to turn it back to False for Runs 1.2i and 2.0i, by also using `config/imsim/processCcd.py` as Jim mentioned above.  So this PR should be all set.  Once approved we'll kindly ask the DM folks to allow a merge?

PR #36, opened 2018-10-25
tickets/DM-16295 - Get cp_pipe etc working for ts8 for bootcamp

mfisherlevine:


python/lsst/obs/lsst/ts8.py
```diff
@@ -187,7 +187,14 @@ def translate_visit(self, md):
...
 
-        return computeVisit(dayObs, seqNum)
+        fullDateTime = datetime.datetime.strptime(dateObs + "+0000", "%Y-%m-%dT%H:%M:%S.%f%z")
+        justTime = datetime.datetime.strptime(dateObs.split('T')[0], "%Y-%m-%d")
+        justTime = justTime.replace(tzinfo=fullDateTime.tzinfo)  # turn into a timedelta obj
+
+        fullDateTime -= ROLLOVERTIME
+        secondsIntoDay = (fullDateTime - justTime).total_seconds()
+
+        return computeVisit(dayObs, secondsIntoDay)
```
TallJimbo: It would probably be good to paste something like the commit message as a code comment here, so it isn't necessary to use `git blame` to understand the sadness that leads to us using `secondsIntoDay` as if it were a sequence number.

python/lsst/obs/lsst/ts8.py
```diff
@@ -51,7 +51,7 @@ def computeVisit(dayObs, seqNum):
 
     try:
         date = datetime.data.fromisoformat(dayObs)
-    except AttributeError:          # requires py 3.7
+    except AttributeError:  # the above line requires py 3.7, raises AttributeError below
```
TallJimbo: I'm afraid I still don't fully understand the second half of this sentence.

policy/ts8/ts8Mapper.yaml
```diff
@@ -4,3 +4,7 @@ calibrations:
     - detector
     - dayObs
     - seqNum
+
+datasets:
+  eotest_gain:  # set of amplifier gains, as measured by eotest
+    template: gain/%(run)s/gain_%(run)s-det%(detector)03d.pickle
```
TallJimbo: Should the rest of the `eotest_gain` dataset definition be moved to `ts8Mapper.yaml`, too?  Or is it actually relevant for `LsstCamMapper` as well?

python/lsst/obs/lsst/ingest.py
```diff
@@ -7,8 +7,11 @@
 
 EXTENSIONS = ["fits", "gz", "fz"]  # Filename extensions to strip off
 
-__all__ = ["LsstCamParseTask"]
+# This is needed elsewhere and should only be defined in one place
+# roll over at 8am UTC
+ROLLOVERTIME = datetime.timedelta(hours=8)
```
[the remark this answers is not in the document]
timj: Some of this will move into the astro_metadata_translator implementation which already uses astropy.time everywhere.

python/lsst/obs/lsst/ts8.py
timj: What does visit really mean here? This definition of visit seems to give you a different visit for every exposure. I don't think we have sufficient metadata to handle two exposures per visit do we?

config/ts8/ingest.py
```diff
@@ -2,46 +2,31 @@
...
+# what used to be config.parse.translators = {
+del config.parse.translators['raftName']
+del config.parse.translators['snap']
+config.parse.translators['dayObs'] = 'translate_dayObs'
+config.parse.translators['filter'] = 'translate_filter'
+config.parse.translators['visit'] = 'translate_visit'
+
+# what used to be config.parse.defaults = {
+config.parse.defaults['imageType'] = "UNKNOWN"
+
```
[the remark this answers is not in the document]
mfisherlevine: As discussed OOB, this is going to stay as-is, because it's where people expect to find this stuff, and it's going to die after Gen3.

python/lsst/obs/lsst/ingest.py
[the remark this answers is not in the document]
mfisherlevine: As discussed OOB, this will be left as-is for now as `datetime` appears in many places, but a ticket will be filed to make sure this is looked at during package-wide review.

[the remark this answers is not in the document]
TallJimbo: Trying to make Gen2 ingest share logic with `astro_metadata_translator` seems like it's going to be both tricky and valuable for this package in particular (IIRC, in other obs packages so far, `astro_metadata_translator` unifies Gen3 ingest and VisitInfo extraction, but not Gen2 ingest).  We may have to end up making one or both a bit uglier temporarily (i.e. lots of otherwise-unnecessary indirection) to make sure each bit of tricky logic is only defined in one place.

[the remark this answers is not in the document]
timj: Yes. I considered that a little when I was looking at obs_subaru but decided I didn't want to get into rewriting the gen2 metadata translator as well. We can revisit that if it looks like we are going to be supporting gen2 and gen3 for some time.

python/lsst/obs/lsst/ts8.py
mfisherlevine: `visit` here is really just an identifier for the exposure. Thankfully this is just test stand data, so there will never be two exposures per visit 🙂

mfisherlevine: Have moved the commit message not as a code comment, but better, into the docstring 🙂

policy/ts8/ts8Mapper.yaml
mfisherlevine: No, I just needed to redefine the template as there are no `raftName`s in ts8 as it's a single raft.

python/lsst/obs/lsst/ingest.py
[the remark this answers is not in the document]
TallJimbo: I suspect we should try to do that unification for obs_lsst only, because its the only one where the Gen2 translators are still being modified frequently (well, maybe obs_decam falls into that category, too; I'm not sure I'd know if it did).

PR #37, opened 2018-11-05
tickets/DM-16390 - Remove seqNum from obs_lsst/ts8 and prepare obs_lsst for bootcamp

mfisherlevine:


config/ts8/ingest.py
```diff
@@ -5,7 +5,7 @@
 # what used to be config.parse.translation = {
 del config.parse.translation['filter']
 del config.parse.translation['visit']
-config.parse.translation['seqNum'] = 'SEQNUM'
+config.parse.translation['testSeqNum'] = 'SEQNUM'
```
SimonKrughoff: I thought you were removing the sequence number.  Is there a reason to leave this in?

config/ts8/ingest.py
```diff
@@ -24,7 +24,7 @@
 del config.register.columns['raftName']
 del config.register.columns['snap']
 config.register.columns['dayObs'] = 'text'
-config.register.columns['seqNum'] = 'int'
+config.register.columns['testSeqNum'] = 'int'
```
SimonKrughoff: Same as above.

policy/auxTel/auxTelMapper.yaml
```diff
@@ -3,9 +3,24 @@ calibrations:
...
+    template: dark/%(calibDate)s/dark-det%(detector)03d_%(calibDate)s.fits
+
+  flat:
+    obsTimeName: dayObs
+    columns:
+    - dayObs
+    - detector
+    - filter
+    template: flat/%(filter)s/%(calibDate)s/flat_%(filter)s-det%(detector)03d_%(calibDate)s.fits
+
```
SimonKrughoff: Why are there changes to auxTel mappers in a ticket for updating ts8?

python/lsst/obs/lsst/ts8.py
```diff
@@ -135,24 +132,60 @@ def translate_detectorName(self, md):
...
-            'E2V-CCD250-182': 1,  # S01
-            'E2V-CCD250-175': 2,  # S02
-            'E2V-CCD250-167': 3,  # S10
-            'E2V-CCD250-195': 4,  # S11
-            'E2V-CCD250-201': 5,  # S12
-            'E2V-CCD250-222': 6,  # S20
-            'E2V-CCD250-213': 7,  # S21
-            'E2V-CCD250-177': 8,  # S22
-        }[serial]
+        # this seem to be appended at more or less at random and breaks the mapping dict
```
SimonKrughoff: I know it's just a comment, but can you clean up the grammar here please?

config/ts8/ingest.py
mfisherlevine: We might want it at a later date (I can imagine someone caring what order the biases for a given test were taken, for example, given the amount of badness in some of them), so best to leave it to save having to reingest later if we want it.

config/ts8/ingest.py
mfisherlevine: Same reply 😉

policy/auxTel/auxTelMapper.yaml
mfisherlevine: Good question 😐 The answer though, is that ts8 inherits from AuxTel.

python/lsst/obs/lsst/ts8.py
mfisherlevine: Sure :)

config/ts8/ingest.py
RobertLuptonTheGood: Is `run` included too?  I think that `(run, seqNum)` is a valid identifier

mfisherlevine: Run is included, but `(run, seqNum)` isn't a valid identifier (you'd need at least `testType` as well)

RobertLuptonTheGood:
Right; does `(run, seqNum, testType)` work?  It should.

mfisherlevine:
> Right; does `(run, seqNum, testType)` work? It should.

I just tried and it doesn't. For anyone reading this in future, the above would actually need to be `(run, testSeqNum, testType)` (as seqNum was replaced with testSeqNum), but I'm going to declare this beyond the scope and merge as I don't think this is a regression really, and it is likely nobody will ever want this, and Gen3 should obviate this whole thing anyway, and debugging this stuff can be *very* time consuming.

PR #41, opened 2018-11-21
set gains, read noise and crosstalk for imsim

jchiang87:


cwwalter:
I'm not very familiar with this package yet.  Is the reason there are only additions to files in this PRM that the imSim info used to be somewhere else?

cwwalter:
Another quick question: I see there is raft offset info in raft.yaml.

So, I'm curious: in https://github.com/LSSTDESC/imSim/issues/156 we discuss how we are still getting geometry info for sims_galSimInterface from obs_lsstSim. Is this info duplicative / in sync with that info?  I'm just wondering how it might interface with the sensor positions there which eventually come from the segmentation.txt file if it is the same info. Is it nothing to worry about?

jchiang87:
When imSim was using the `LsstCamMapper`, the gain and read noise info was coming from the files in `policy/lsstCam`.   I added the corresponding files to `policy/imsim` with the gain and read noise set, and added `ccdData.yaml` for the crosstalk values.  I included a copy of `rafts.yaml` since it was also in `policy/phosim`, but maybe just having `policy/rafts.yaml` is sufficient.  In any case, I don't think anything is being duplicated vis-a-vis obs_lsstSim, and segmentation.txt is irrelevant at this point.

cwwalter:
Thanks. 

Structure and values looks good to me, but I haven't really looked at this package before. Probably good to get a review from a real expert.  

Thinking ahead: If we have comCam or different FP configurations we would have different directories with other names (like imsim-comcam)?

Happy Thanksgiving all!

heather999:
After consulting with Robert, Jim did some additional checking.  This is good to go, now rebased, and merging to master.

RobertLuptonTheGood:
Re "Thinking ahead: If we have comCam or different FP configurations we would have different directories with other names (like imsim-comcam)?"
Providing you only have one camera configuration that is valid at any one time (so one imsim, one phosim, one comCam, one lsstCam, ...) we plan to handle this via versioning the information like any other calibration product (e.g. a flat field).   *But* that requires butler Gen3

PR #51, opened 2018-12-20
rationalize handling of doCrosstalk for imsim and phosim

heather999:
Handles both processCcd and singleFrameDriver
Tested both processCcd and singleFrameDriver using DESC imsim Run1.2i and Run1.2p repos:
`processCcd.py <pathtorepo> --rerun checkConfig --show config=isr.doCrosstalk`
This change should only impact imsim and phosim processing.

### DIFF
--- a/config/imsim/processCcd.py
+++ b/config/imsim/processCcd.py
@@ -20,4 +20,4 @@
 # the GNU General Public License along with this program.  If not,
 # see <http://www.lsstcorp.org/LegalNotices/>.
 
-config.isr.doCrosstalk=False
+config.isr.doCrosstalk=True
--- a/config/imsim/singleFrameDriver.py
+++ b/config/imsim/singleFrameDriver.py
@@ -1,0 +1,1 @@
+config.processCcd.isr.doCrosstalk=False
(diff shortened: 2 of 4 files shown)


johannct:
my understanding is that singleFrameDriver is needed in the subdir to load the isr.py present in the subdir. Ok for setting True while we are not yet in master. I am not convinced that duplicating config allows to avoid further confusion. If I do not see any override in phosim, I conclude that we use the default; if I see an override, I tend to conclude that the default is different from the override value.....

heather999:
My concern is what happens if later in master `config/isr.py` no longer has that `isr.doCrosstalk=True`? Then we'll revert to doCrosstalk=False, as that seems to be the default outside of `obs_lsst`.  That's why I wanted to set it explicitly in `config/phosim`.  So we're not surprised later. But I can remove it for now in the name of progress!

timj:
@heather999 I have rebased this pull request. Are you intending to merge it?

heather999:
@jchiang87 I want to close out this PR. One last question though - while for phosim doCrossTalk=True consistently makes sense for imsim, it depends on what it is we're processing, whether it's Run1.2i, 2.0i (false) or 2.1i (true) - right? So should I leave it set to True as the default for imsim since I'm expecting from 2.1i onward we'd want xtalk on?

PR #54, opened 2019-01-10
DM-17095: Speed up tests

timj:
Creating a butler seems to take "forever" and we are doing it every test. This PR creates a butler ever test class. Tests now take 3 minutes rather than 15 minutes on my laptop.

### DIFF
--- a/tests/test_auxTel.py
+++ b/tests/test_auxTel.py
@@ -24,28 +24,15 @@ import sys
 import unittest
 
 import lsst.utils.tests
-from lsst.utils import getPackageDir
 from lsst.afw.geom import arcseconds, Extent2I
 import lsst.afw.image
 import lsst.obs.base.tests
 
+from lsst.obs.lsst.testHelper import ObsLsstButlerTests
 
-class TestAuxTel(lsst.obs.base.tests.ObsTests, lsst.utils.tests.TestCase):
-    @classmethod
-    def tearDownClass(cls):
-        cls._mapper._LsstCamMapper__clearCache()
-        del cls._mapper
-        del cls._butler
 
-    @classmethod
-    def setUpClass(cls):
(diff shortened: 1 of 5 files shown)


tests/test_auxTel.py
```diff
@@ -24,28 +24,15 @@
 import unittest
 
 import lsst.utils.tests
-from lsst.utils import getPackageDir
 from lsst.afw.geom import arcseconds, Extent2I
 import lsst.afw.image
 import lsst.obs.base.tests
 
+from obsLsstTestHelper import ObsLsstButlerTests
```
SimonKrughoff: Should this be explicitly relative?

tests/test_auxTel.py
```diff
@@ -24,28 +24,15 @@
...
 
-class TestAuxTel(lsst.obs.base.tests.ObsTests, lsst.utils.tests.TestCase):
```
SimonKrughoff: There is something wrong with this diff.  The PR is asking to go to master, but these class level tearDown and setUp do not exist on master?  Was it branched from another ticket, and if so, will that ticket get merged first?

SimonKrughoff: Oh never mind.  I see that you did it in two pieces.  I think it would be clearer if these two commits were squashed.

timj: Well, one commit was the speed up commit. The second commit is the refactoring to pull the code into a helper class.

tests/test_auxTel.py
timj: Only if we add a `__init__.py` I think.

tests/test_auxTel.py
SimonKrughoff: I'm happy with either way.  It confused me because I read the commits in the wrong order.

PR #57, opened 2019-01-15
DM-17190: obs_lsst needs to dynamically switch between ts8 ITL or e2v sensors

hsinfang:
hack hack hack...


policy/SConscript
```diff
@@ -17,7 +17,7 @@ libpathstr = lsst.sconsUtils.utils.libraryLoaderEnvironment()
 # We always run these commands with an explicit python rather than relying on the shebang
 python = "{} python".format(libpathstr)
 
-for camera in ["auxTel", "lsstCam", "imsim", "phosim", "ts8", "ts8itl"]:
+for camera in ["auxTel", "lsstCam", "imsim", "phosim", "ts8e2v", "ts8itl"]:
```
RobertLuptonTheGood: I don't think we want to do this.  The need for two mappers is purely due to missing butler functionality which Gen3 will address.  At the worst, we could make `ts8e2v` an alias for `ts8`.

I only added `ts8itl` as a temporary workaround.

mfisherlevine: Right, absolutely, but I think while we wait for Gen3 this is necessary, and there is no reason that the nominal case should be one or the other, so the vanilla ts8 should disappear, and be replaced with two equal cases, `ts8e2v` and `ts8itl`. Unless there are reasons for not doing it like this in the meantime I think this is how it should proceed, and we can fix it up once we have version-able everything come Gen3.

tests/test_ts8.py
```diff
@@ -33,7 +33,7 @@
 
 
 class TestTs8(lsst.obs.base.tests.ObsTests, ObsLsstButlerTests):
-    instrumentDir = "ts8"
+    instrumentDir = "ts8e2v"
```
mfisherlevine: Would be great to see the tests extended to both cameras, but given that the tests in here are minimal, I won't block the merge on that, especially as this is under heavy development elsewhere.

hsinfang: Agree. DM-17414 is filed to capture such future work.

PR #58, opened 2019-01-15
DM-17228: Switch gen2 translators to gen3 astro_metadata_translator

timj:
* Add tests for ParseTask metadata extraction
* Replace Gen2 translator logic with calls to astro_metadata_translator translators
* Replace ccd exposure ID calculation with astro_metadata_translator logic


python/lsst/obs/lsst/translators/ts8.py
```diff
@@ -88,6 +100,94 @@ def can_translate(cls, header, filename=None):
...
+            Date of observation in FITS ISO format.
+        seqnum : `int`, unused
+            Sequence number. Ignored.
+
+        Returns
+        -------
+        exposure_id : `int`
+            Exposure ID.
+        """
+        # There is worry that seconds are too course so use 10th of second
```
mfisherlevine: *coarse 😛

python/lsst/obs/lsst/ts8.py
```diff
@@ -348,7 +142,7 @@ def translate_testSeqNum(self, md):
 
         Parameters
         ----------
-        md : `lsst.daf.base.PropertyList or PropertySet`
+        md : `lsst.daf.base.PropertyList` or `~lsst.daf.base.PropertySet`
```
mfisherlevine: As per previous comments about `~`s

tests/test_auxTel.py
```diff
@@ -43,7 +43,7 @@ def setUp(self):
         self.setUp_tests(self._butler, self._mapper, dataIds)
 
         ccdExposureId_bits = 32
-        exposureIds = {'raw': 1140013000, 'bias': 1140013000}
+        exposureIds = {'raw': 5700065, 'bias': 20180920000065}
```
mfisherlevine: I see that this is dealing with the new naming scheme, but before they were the same, and now they're different. Is this correcting a mistake from before as well as updating? I mean, that would make sense, as I don't see why the raw and bias would have the same IDs, but just checking I'm following what's going on here.

mfisherlevine: Also, why does the bias ID seem to follow the new date-contact format, but not the raw?

tests/test_imsim.py
```diff
@@ -43,10 +43,10 @@ def setUp(self):
         self.setUp_tests(self._butler, self._mapper, dataIds)
 
         ccdExposureId_bits = 32
-        exposureIds = {'raw': 40919042,
-                       'bias': 40919042,
-                       'dark': 40919042,
-                       'flat': 40919042
+        exposureIds = {'raw': 2045950042,
```
mfisherlevine: What are these new numbers? I can imagine the 2045 could be from the date-concat as they're simulated in the future, but then I don't understand the rest of the digits or why they're so short.

tests/test_imsim.py
```diff
@@ -43,10 +43,10 @@ def setUp(self):
...
+                       'bias': 2045950042,
+                       'dark': 2045950042,
+                       'flat': 2045950042
```
mfisherlevine: and I know it was like this before, but does it really make sense that these are all the same everywhere?

config/ingest.py
```diff
@@ -3,26 +3,25 @@
 config.parse.retarget(LsstCamParseTask)
 
 config.parse.translation = {
-    'expTime': 'EXPTIME',
-    'object': 'OBJECT',
-    'imageType': 'IMGTYPE',
     'testType': 'TESTTYPE',
```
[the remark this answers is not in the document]
timj: There is no equivalent in `ObservationInfo` for this information so it has to be handled the old way.

config/ingest.py
```diff
@@ -3,26 +3,25 @@
...
+    'dateObs': 'translate_dateObs',
+    'run': 'translate_run',
+    'visit': 'translate_visit',
     'wavelength': 'translate_wavelength',
     'raftName': 'translate_raftName',
     'detectorName': 'translate_detectorName',
     'detector': 'translate_detector',
     'snap': 'translate_snap',
 }
 config.parse.defaults = {
```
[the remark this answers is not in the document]
timj: I think I prefer empty so that it forces the issue and you know what state you have.

python/lsst/obs/lsst/ingest.py
```diff
@@ -44,15 +83,15 @@ def translate_wavelength(self, md):
 
         Parameters
         ----------
-        md : `lsst.daf.base.PropertyList or PropertySet`
+        md : `~lsst.daf.base.PropertyList` or `~lsst.daf.base.PropertySet`
```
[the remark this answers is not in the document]
timj: The `~` tells sphinx to render it as `PropertyList` rather than `lsst.daf.base.PropertyList` whilst also making sure it links to the full name.

python/lsst/obs/lsst/ingest.py
```diff
@@ -23,13 +25,50 @@ class LsstCamParseTask(ParseTask):
...
+
+        The translator methods receive the header metadata and should return the
+        appropriate value, or None if the value cannot be determined.
+
+        Parameters
+        ----------
+        md : `lsst.daf.base.PropertyList`
+            FITS header
+        info : `dict`, optional
+            File properties, to be supplemented
```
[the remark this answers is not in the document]
timj: I copied this text from the base class 😄 I'll tweak here.

python/lsst/obs/lsst/ingest.py
```diff
@@ -23,13 +25,50 @@ class LsstCamParseTask(ParseTask):
...
+        super().__init__(config, *args, **kwargs)
 
+        self.observationInfo = None
         if self.camera is None:
             self.camera = self._cameraClass()
 
+    def getInfoFromMetadata(self, md, info=None):
+        """Attempt to pull the desired information out of the header.
+
+        Notes
```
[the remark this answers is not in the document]
timj: This structure came from the base class but I'll reorganize.

python/lsst/obs/lsst/translators/auxTel.py
```diff
@@ -44,6 +44,12 @@
 # since the headers have not historically been reliable
 TSTART = Time("2019-06-01T00:00", format="isot", scale="utc")
 
+# Define the sensor and group name for AuxTel globally so that it can be used
+# in multiple places. There is no raft but for consistency with other LSST
+# cameras we define one.
+_DETECTOR_GROUP_NAME = "RXX"
+_DETECTOR_NAME = "S00"
```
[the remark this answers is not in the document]
timj: The only constraint is that they have to match the names used in the camera policy YAML files. I chose names to match what other people had already chosen. I'd like to defer poking this on this ticket. We can have the discussion with a wider audience later.

tests/test_auxTel.py
timj: Because I'm not recreating the butler repository in the git repository. One of those IDs is calculated dynamically (the bias) and the other is previously stored in the butler repo.  To unify things I need to talk to @SimonKrughoff to work out how he made the repository in the first place. I can do that on a separate ticket.

tests/test_imsim.py
timj: ImSim exposure ID comes from the OBSID header (204595 in this case). The 0042 is the padding of the detector number. This is the same exposure ID as was used previously but without the multiply by 200.

tests/test_imsim.py
timj: No. It does not make sense that one exposure_id is the same for a bias, a dark, and a flat. This probably means that the headers aren't changing when new files are created and the translators need to work harder to make things unique. @jchiang87 should probably comment here (but again we can fix this in a different ticket).

python/lsst/obs/lsst/ingest.py
```diff
@@ -66,50 +105,45 @@ def translate_dateObs(self, md):
...
 
         Returns
         -------
         dateObs : `str`
-            The day that the data was taken, e.g. 2018-08-20T21:56:24.608
+            The date that the data was taken, e.g. 2018-08-20T21:56:24.608
         """
-        return self.__fixDateObs(md.get("DATE-OBS"))
+        dateObs = self.observationInfo.datetime_begin
+        dateObs.format = "isot"
```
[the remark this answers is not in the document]
timj: Actually, it will work because I don't even read the DATE-OBS header. The new translator uses the MJD-OBS field. Marking as resolved.

python/lsst/obs/lsst/ingest.py
```diff
@@ -66,50 +105,45 @@ def translate_dateObs(self, md):
...
-        dateObs = self.__fixDateObs(md.get("DATE-OBS"))
-
-        d = datetime.datetime.strptime(dateObs + "+0000", "%Y-%m-%dT%H:%M:%S.%f%z")
-        d -= ROLLOVERTIME
-        dayObs = d.strftime("%Y-%m-%d")
-
-        return dayObs
+        dateObs = self.observationInfo.datetime_begin
+        dateObs -= MDROLLOVERTIME
+        dateObs.format = "iso"
```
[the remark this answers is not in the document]
timj: Don't worry about it. The AuxTel translator uses the MJD-OBS field and not the DATE-OBS field.

tests/test_imsim.py
jchiang87: Are these data from auxtel or TS8?   For TS8 data, we have had run numbers, which are unique to an acquisition sequence that may contain multiple types of acquisitions (biases, darks, flats, etc.).  For `OBSID`, I can imagine that the code that writes that keyword just fills it with the `RUNNUM` value, hence the same value for these different types of exposures.  I have no idea what they're doing with auxtel data, but I suspect it is something similar.  For real data (i.e., not imSim or PhoSim), all of this is done by the CCS software.

timj: This is ImSim.

jchiang87: Sorry, was misled by the earlier comments.  Which imsim files are these numbers derived from?

jchiang87: https://github.com/lsst/obs_lsst/blob/master/data/input/imsim/raw/204595/R11/00204595-R11-S20-det042-000.fits  looks like a raw imsim sky exposure file, but with all of the pixel data set to zero (presumably to make the file smaller).  That's fine, and the `OBSID = 204595` looks like it corresponds to an actual visit from one of our simulation runs.   The files in https://github.com/lsst/obs_lsst/tree/master/data/input/imsim/CALIB  are the outputs from running `construct[Bias,Dark,Flat].py` on imsim raw calibration frames.   Those headers have the unique `OBSID`s for the frames listed, i.e., for the bias file
```
HIERARCH CALIB_INPUT_0 = '(3010000,)'                                           
HIERARCH CALIB_INPUT_1 = '(3010003,)'                                           
HIERARCH CALIB_INPUT_2 = '(3010006,)'                                           
HIERARCH CALIB_INPUT_3 = '(3010009,)'          
```
I don't see `OBSID` in any of the `CALIB` file headers, so I'm not sure what on the imsim side needs to be done any differently.

timj: Thanks. That explains it. I think calibration products have some issues. I think they are meant to use a different code path for visitID calculation but I'm not an expert on that (we generate special Ids for coadds on sky tracts but I don't know what happens for calibrations). I also really would like more complete headers in our calibration products.

python/lsst/obs/lsst/ingest.py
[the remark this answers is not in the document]
mfisherlevine: Oh, that's really nice - I'm learning! (Sorry this is duplicated everywhere!)

python/lsst/obs/lsst/ingest.py
```diff
@@ -44,15 +83,15 @@ def translate_wavelength(self, md):
...
+        md : `~lsst.daf.base.PropertyList` or `~lsst.daf.base.PropertySet`
             image metadata
 
         Returns
         -------
         wavelength : `int`
             The recorded wavelength as an int
         """
-        raw_wl = md.get("MONOWL")
+        raw_wl = md.getScalar("MONOWL")
```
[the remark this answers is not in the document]
mfisherlevine: Yes, I think I am. Assuming this returns the value of `-666` if the key isn't found then I think the behaviour will be the same.

python/lsst/obs/lsst/ingest.py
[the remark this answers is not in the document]
mfisherlevine: Oh, I could be mistaken, but I thought from memory that this was dealing with a weird acq system bug that meant that sometimes the formatting was messed up, with things like the hour not always being zero-padded in the string, or something like that...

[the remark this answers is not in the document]
mfisherlevine: Looking at the code though, that doesn't seem to be the case, so I guess this is fine, but I'm not sure where the thing I'm thinking of has gone, so I guess we'll just wait for an ingest to fail!

python/lsst/obs/lsst/translators/auxTel.py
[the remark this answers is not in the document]
mfisherlevine: Sure, that sounds fine, no reason not to punt on this.

tests/test_auxTel.py
mfisherlevine: OK, sounds good

python/lsst/obs/lsst/ingest.py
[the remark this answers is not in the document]
timj: Previously the translators read DATE-OBS from the header and DATE-OBS used to have a trailing `(UTC)` that needed to be stripped out. The new translator uses MJD-OBS which doesn't have that problem.

PR #64, opened 2019-01-25
DM-17375: Update all the license files and copyright

timj:
* Add GPL and copyright file.
* Update preambles.
* Explicitly state that the translators are BSD 3-clause.


ktlim:
Do we really need preamble in config files?  They're persisted data (even though they happen to be Python syntax).

timj:
@ktlim Regarding config files, they are code, sometimes a lot of code. I can back out that final commit if you prefer.

PR #65, opened 2019-01-25
DM-17299: Deal with arbitrary and untested scripts

SimonKrughoff:
This moves things around so they are slightly more logical locations.  It also adds a test for one of the scripts.


bin.src/phosimToRafts.py
```diff
@@ -102,10 +131,14 @@
...
+        raftSerial = f"LCA-11021_RTM-{raftId:03d}"  # won't be deterministic in reality!
         raftId += 1
 
         if raftName in ("R00", "R40", "R04", "R44"):
             continue
-        with open("%s.yaml" % raftName, "w") as fd:
+        if args.output_dir:
+            prefix = args.output_dir
+        else:
+            prefix = ""
```
timj: I prefer to be explicit here and use `os.path.curdir`

tests/test_pho2rafts.py
```diff
@@ -0,0 +1,65 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
+from __future__ import print_function
```
timj: This is not needed. Delete.

tests/test_pho2rafts.py
```diff
@@ -0,0 +1,65 @@
...
+        # Read file produced by script above
+        with open(os.path.join(TESTDIR, 'R11.yaml')) as fh:
+            doc = yaml.load(fh)
+        # Read file with expected outputs
+        with open(os.path.join(EXPYAMLDIR, 'R11.yaml')) as fh:
+            exp_doc = yaml.load(fh)
+        # The test data only have the S20 chip
+        for key in list(exp_doc['R11']['amplifiers']):
+            if key != 'S20':
+                _ = exp_doc['R11']['amplifiers'].pop(key, None)  # NOQA F841
```
timj: The `pop` here is kind of cute but I'm not sure why it's necessary. You know that the `key` must exist because you are looping over the keys, so this should be fine:
```python
if key != 'S20':
    del exp_doc['R11']['amplifiers'][key]
```
The noqa is trying to tell you something.

tests/test_pho2rafts.py
```diff
@@ -0,0 +1,65 @@
...
+                              args=[DATADIR, "--visit", "204595", "--output_dir", TESTDIR],
+                              msg="phosimToRafts.py failed")
+        # Read file produced by script above
+        with open(os.path.join(TESTDIR, 'R11.yaml')) as fh:
+            doc = yaml.load(fh)
+        # Read file with expected outputs
+        with open(os.path.join(EXPYAMLDIR, 'R11.yaml')) as fh:
+            exp_doc = yaml.load(fh)
+        # The test data only have the S20 chip
+        for key in list(exp_doc['R11']['amplifiers']):
```
timj: Maybe add comment here that you have to use `list()` because you are deleting from the dict.

tests/test_pho2rafts.py
```diff
@@ -0,0 +1,65 @@
...
+            if key != 'S20':
+                _ = exp_doc['R11']['amplifiers'].pop(key, None)  # NOQA F841
+        # Can't compare since the serial depends on how many rafts are in the
+        # input repository.  The test repository only has 1.
+        _ = doc['R11'].pop('raftSerial', None)  # NOQA F841
```
timj:
```python
del doc['R11']['raftSerial']
del exp_doc['R11']['raftSerial']
```
or even
```python
for d in (doc, exp_doc):
    del d['R11']['raftSerial']
```

tests/test_pho2rafts.py
SimonKrughoff: Good point.  Done.

tests/test_pho2rafts.py
SimonKrughoff: Much nicer.

tests/test_pho2rafts.py
```diff
@@ -0,0 +1,65 @@
+#
+# LSST Data Management System
+# Copyright 2012-2017 LSST Corporation.
```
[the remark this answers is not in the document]
SimonKrughoff: Good catch.  Obviously I cargo culted the header from the wrong place.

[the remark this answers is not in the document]
timj: I think when you started the work they all did this. #64 changed them all.

PR #43, opened 2018-11-28
U/cslage/ucd cam

craiglagegit:
Based on what I learned at the boot camp, I've created a ucdCam in obs_lsst so I can ingest  fits files from the UC Davis camera into the LSST stack.  I created a branch in obs_lsst called u/cslage/ucdCam.  I basically just copied the ts8 files and made the necessary changes to parse the UC Davis headers.  It's all working for both E2V and ITL fits files, but I wasn't able to push it into lsst/obs_lsst because I don't have permission.  Based on input from Simon Krughoff, I created a fork in craiglagegit/obs_lsst.  I want to merge this into the master branch so other people (like Andrew Bradshaw) can use it.  There is an example notebook at 

https://github.com/craiglagegit/obs_lsst/blob/u/cslage/ucdCam/examples/UCD_Ingest_27Nov18.ipynb

that shows the testing I've done.  It's not completely necessary to merge it into the master branch,  we could continue to pull it from the fork, but I'd like to do this so it gets carried along in future updates.  Let me know if I need to do more to merge it in.  Thanks!

PR #66, opened 2019-01-28
DM-17437: Add UCDavis test camera

timj:
This is  #43 rewritten to use the new translation infrastructure.  After discussions with @RobertLuptonTheGood and @mfisherlevine we have decided that the three detectors should be named R00_S00, R01_S00, and R02_S00 rather than S00, S01, S02.

cc/ @craiglagegit

### DIFF
--- a/policy/ucd/R00.yaml
+++ b/policy/ucd/R00.yaml
@@ -6,19 +6,19 @@ R00 :
 
   amplifiers :
     S00 :
-      C00 : { gain : 1.700, readNoise : 7.00 }
-      C01 : { gain : 1.700, readNoise : 7.00 }
-      C02 : { gain : 1.700, readNoise : 7.00 }
-      C03 : { gain : 1.700, readNoise : 7.00 }
-      C04 : { gain : 1.700, readNoise : 7.00 }
-      C05 : { gain : 1.700, readNoise : 7.00 }
-      C06 : { gain : 1.700, readNoise : 7.00 }
-      C07 : { gain : 1.700, readNoise : 7.00 }
-      C10 : { gain : 1.700, readNoise : 7.00 }
-      C11 : { gain : 1.700, readNoise : 7.00 }
-      C12 : { gain : 1.700, readNoise : 7.00 }
-      C13 : { gain : 1.700, readNoise : 7.00 }
-      C14 : { gain : 1.700, readNoise : 7.00 }
-      C15 : { gain : 1.700, readNoise : 7.00 }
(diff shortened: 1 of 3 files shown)


policy/SConscript
```diff
@@ -30,6 +30,6 @@ for camera in ["auxTel", "lsstCam", "imsim", "phosim", "ts8e2v", "ts8itl"]:
     for d in dirList:
         for f in glob.glob(os.path.join(d, "*.yaml")):
             env.Depends(commandInst, f)
-
```
RobertLuptonTheGood: Please revert whitespace changes

policy/cameraHeader.yaml
```diff
@@ -248,6 +248,7 @@ CCD_ITL : &CCD_ITL
         <<: *C07_ITL
         hdu : 16
         ixy : [0, 1]
+
```
RobertLuptonTheGood: Please revert whitespace changes

policy/ucd/R00.yaml
```diff
@@ -0,0 +1,24 @@
+R00 :
+  detectorType : E2V_UCD
+  raftSerial : LCA-11021_RTM-E2V-00
+  ccdSerials :
+    S00 : E2V-CCD250-112-04
+
+  amplifiers :
+    S00 :
+      C00 : { gain : 1.700, readNoise : 7.00 }
```
RobertLuptonTheGood: Do we have better gain/readNoise values?  Applies to all three CCDs

policy/SConscript
timj: Why? We are not supposed to have trailing whitespace in python files.

policy/cameraHeader.yaml
timj: Why do we want trailing whitespace in YAML files? It serves no purpose (and it's really hard for me to keep it around based on my editor settings and git moaning a lot during rebases).

policy/ucd/R00.yaml
timj: That's for @craiglagegit to answer.

craiglagegit: I assumed those were just placeholders, since the numbers there don't seem to apply to any particular system.  However, the answer is yes, I have better gain/readNoise numbers for all three detectors.  Should I go ahead and put those in?

RobertLuptonTheGood: Yes please.

craiglagegit: OK, new numbers pushed.

PR #70, opened 2019-01-29
DM-17530: Add support for BNL TS3 data

timj:
* Basic translator so far.
* Need to add mappers and camera definition files.

@mfisherlevine should we use the #66 approach of having R00 be E2V data and R01 be ITL data? Do you have a list of all the detector serial numbers to be supported? Do we have a raft per sensor to make the camera yaml easier?


timj:
@mfisherlevine take a look at this and see what you think.  I can ingest those two TS3 files now without a problem. I need to add nulled out versions to the local obs_lsst butler repo so I can run the tests.

python/lsst/obs/lsst/ts3.py
```diff
@@ -0,0 +1,87 @@
...
+
+        Returns
+        -------
+        seqNum : `int`
+            The seqNum, with a default value of ``0`` if required.
+        """
+        try:
+            seqNum = md.getScalar("SEQNUM")
+        except KeyError:
+            seqNum = 0
```
mfisherlevine: This feel dangerous, as it could return the same value for a bunch of things, all of them looking valid. Would -1 break things? I don't really know where else this is used (likely nowhere, but I'm unsure)

python/lsst/obs/lsst/translators/ts8.py
```diff
@@ -143,7 +143,7 @@ def to_instrument(self):
         instrume : `str`
             Name of the test stand.
         """
-        return f"LSST-TS8"
+        return "LSST-TS8"
```
mfisherlevine: Seems unrelated, but I don't actually care.

python/lsst/obs/lsst/translators/ts3.py
```diff
@@ -0,0 +1,242 @@
...
+        filename : `str`, optional
+            Name of file being translated.
+
+        Returns
+        -------
+        can : `bool`
+            `True` if the header is recognized by this class. `False`
+            otherwise.
+        """
+        return cls.can_translate_with_options(header, {"TSTAND": "BNL-TS3-2-Janeway"}, filename=filename)
```
mfisherlevine: Is that a default value? If so, I'm not sure it's appropriate. If not then it might well be fine.

python/lsst/obs/lsst/translators/lsstsim.py
```diff
@@ -162,6 +162,7 @@ def compute_detector_info_from_serial(cls, detector_serial):
             info = serial_mapping[detector_serial]
         else:
             log.warning(f"Unable to determine detector information from detector serial {detector_serial}")
+            info = None
```
mfisherlevine: Again, unrelated but still don't care :)

python/lsst/obs/lsst/translators/ts8.py
timj: Yes, but I saw something that needed to be fixed so I fixed it rather than creating a new ticket and a new PR so that I could remove and `f`.

python/lsst/obs/lsst/ts3.py
timj: This is exactly what TS8 does. As far as I can tell it will always return 0 because SEQNUM isn't in TS3 files and you shouldn't really use it... If you like I can remove the entire entry and not use seqNum at all for TS3.

python/lsst/obs/lsst/translators/ts3.py
timj: I only have two files and they both have that value for TSTAND.  If that TSTAND value changes over the history of TS3 then you'll find out pretty easily. If you have a bunch of data lying around you it's pretty easy to run `translate_header.py -p lsst.obs.lsst.translators /bunch/of/ts3/data` and see which files it has problems with.

timj: That's not a default value. It's the value that TSTAND header must have in order to decide that the file represents TS3 data. The translation system tries to work out what type of data it has been given.

python/lsst/obs/lsst/ts3.py
mfisherlevine: No no, I think this is fine.

mfisherlevine:
OK, looks good. Thanks Tim!

tests/test_ts3.py
```diff
@@ -0,0 +1,159 @@
...
+import unittest
+
+import lsst.utils.tests
+from lsst.geom import arcseconds, Extent2I
+import lsst.afw.image
+
+from lsst.obs.lsst.testHelper import ObsLsstButlerTests, ObsLsstObsBaseOverrides
+
+
+class TestTs8(ObsLsstObsBaseOverrides, ObsLsstButlerTests):
```
mfisherlevine: Should the class name be changed to `TestTs3`?

tests/test_ts3.py
```diff
@@ -0,0 +1,159 @@
...
+                   }
+        self.setUp_tests(self._butler, self._mapper, dataIds)
+
+        ccdExposureId_bits = 36
+        exposureIds = {'raw': 201607220607067071}
+        filters = {'raw': '550CutOn'}
+        exptimes = {'raw': 30.611}
+        detectorIds = {'raw': 71}
+        detector_names = {'raw': 'R071_S00'}
+        detector_serials = {'raw': 'ITL-3800C-098'}
```
mfisherlevine: Nice choice of chips. Is that a coincidence, or was that deliberate?

python/lsst/obs/lsst/translators/ts3.py
```diff
@@ -19,18 +19,19 @@
...
-# in multiple places. There is only a single sensor so define a
-# raft for consistency with other LSST cameras.
-_DETECTOR_GROUP_NAME = "RXX"
+# There is only a single sensor at a time so define a
+# fixed sensor name
 _DETECTOR_NAME = "S00"
 
 
-class LsstTS3Translator(StubTranslator):
+class LsstTS3Translator(LsstSimTranslator):
```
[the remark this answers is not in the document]
timj: It's telling you that imsim and phosim share infrastructure that is also now needed by TS3 (and TS8 it turns out) and that at this point the parent class name is wrong.

tests/test_ts3.py
timj: You chose them (this is zeroed out data from the testdata repository).

PR #73, opened 2019-01-31
removing config/lsstCam.py and replacing with config/isr.py

heather999:
As discussed in #dm-lsstcam, `isr.py` replaces `lsstCam.py`: [https://lsstc.slack.com/archives/CBE964PR8/p1541704840055600](https://lsstc.slack.com/archives/CBE964PR8/p1541704840055600)  We discussed this a couple of months ago @mfisherlevine and @czwa but I'll ping both of you anyway to make sure we should go ahead and merge.


heather999:
I"m going to forge ahead and merge this in today - unless I hear something by 1 PM ET :)

mfisherlevine:
This PR broke things. The contents of the file that was replaced were not the same, and now config options aren't found. I don't know how many sub-obs_packages are affected, but `ts8` certainly is for calib construction.

RobertLuptonTheGood:
Did you run the `ci_lsst` scripts?  Or is just biases not enough to show the problem?  We should add a more complete set of calibs to the test data

mfisherlevine:
I didn't run anything.

mfisherlevine:
I only just saw that a review was requested from me, as GitHub emails go to a different inbox, and I've been ignoring them so I can focus on stuff. If this had been a Jira ticket & review I wouldn't have missed it, but it's also not the reviewer's job to run tests...

RobertLuptonTheGood:
Sorry, I was asking @heather999, not @mfisherlevine.  There are tests that check out much of the `obs_lsst` functionality, and they should probably have been run before merging.  This will, at some point, probably be automated.

timj:
We are planning to add `ci_lsst` to stack-os-matrix so that you can run it on Jenkins. @SimonKrughoff is looking into it.

SimonKrughoff:
This is certainly the next step on the road to stabilizing `obs_lsst`.  It is next up for me to do.

heather999:
@RobertLuptonTheGood I was under the impression that running `scons` on obs_lsst would effectively run most tests of interest.
@mfisherlevine Since we discussed this back in November, `config/isr.py` was updated to include a couple more items than `config/lsstCam.py` contained.  Here is what lsstCam.py contained before it was removed:
https://github.com/lsst/obs_lsst/blob/566262b766ded1c19e86d4f08be461d692277660/config/lsstCam.py
This is `isr.py` on master: https://github.com/lsst/obs_lsst/blob/master/config/isr.py
Here is the history that shows when two lines were added:
```
config.qa.doThumbnailOss = False
config.qa.doThumbnailFlattened = False
```
https://github.com/lsst/obs_lsst/blame/master/config/isr.py

So my inclination would be to inquire why those lines are there and to determine if they are necessary in all cases.  They are not things that I added, but was under the impression they were added for some good purpose.

mfisherlevine:
Ah, it's not those that broke things, it was the difference between
`config.isr.foo = False` and
`config.foo = False` that got reverted and broke things.
It's OK though, since we worked it out DM-17766 has been filed to fix this.

PR #74, opened 2019-01-31
DM-17292: Improve test coverage

timj:
Significantly improve test coverage, especially in the mappers.


python/lsst/obs/lsst/auxTel.py
```diff
@@ -91,34 +90,6 @@ class AuxTelParseTask(LsstCamParseTask):
     _mapperClass = AuxTelMapper
     _translatorClass = LsstAuxTelTranslator
 
-    def getInfo(self, filename):
```
SimonKrughoff: With this going away, I think the class docstring needs an update.

python/lsst/obs/lsst/lsstCamMapper.py
```diff
@@ -499,23 +508,15 @@ def map_raw_visitInfo(self, *args, **kwargs):
...
-            mat = re.search(r"\[(\d+)\]$", fileName)
-            if mat:
-                hdu = int(mat.group(1))
-                md = readMetadata(fileName, hdu=hdu)
-            else:
-                md = readMetadata(fileName)  # or hdu = INT_MIN; -(1 << 31)
+            md = readMetadata(fileName)  # or hdu = INT_MIN; -(1 << 31)
 
-            return afwImage.VisitInfo(md)
+        return self.MakeRawVisitInfoClass(log=self.log)(md)
```
SimonKrughoff: Hmm.  This is interesting syntax.  I'm assuming the first set of parens is for the `__init__` method and the second is for `__call__`.  Would it be easier to read if the class was constructed on one line and called in the return?

python/lsst/obs/lsst/ts8.py
```diff
@@ -85,35 +83,6 @@ class Ts8ParseTask(LsstCamParseTask):
     _mapperClass = Ts8Mapper
     _translatorClass = LsstTS8Translator
 
-    def getInfo(self, filename):
```
SimonKrughoff: This needs a similar doctoring update.

tests/test_auxTel.py
```diff
@@ -26,12 +26,12 @@
 import lsst.utils.tests
 from lsst.afw.geom import arcseconds, Extent2I
 import lsst.afw.image
-import lsst.obs.base.tests
+import lsst.daf.base
```
SimonKrughoff: It appears that `lsst.daf.base` is unused.

tests/test_parsetask.py
```diff
@@ -94,11 +116,35 @@ def test_parsetask_auxtel_translator(self):
...
+        md["IMGNAME"] = "AT-O-20180816-00008"
+        seqNum = parseTask.translate_seqNum(md)
+        self.assertEqual(seqNum, 8)
+
+        del md["IMGNAME"]
+        md["FILENAME"] = "ats_exp_27_AT_C_20180920_000065.fits"
+        seqNum = parseTask.translate_seqNum(md)
+        self.assertEqual(seqNum, 65)
+
+        # This will issue a warning
```
SimonKrughoff: Did you want to test the warning as you have elsewhere?

tests/test_phosim.py
```diff
@@ -114,6 +113,46 @@ def setUp(self):
...
+        exposureId = self.butler.get('ccdExposureId', dataId={"visit": 1, "detector": 1})
+        self.assertEqual(exposureId, 10001)
+
+        with self.assertRaises(ValueError):
+            self.butler.get('ccdExposureId', dataId={"visit": 1, "detector": 2000})
+
+        with self.assertRaises(KeyError):
+            self.butler.get('ccdExposureId', dataId={"visit": 1})
+
+        exposureId = self.butler.get('ccdExposureId', dataId={"visit": 1, "raftName": "R01",
```
SimonKrughoff: It would read a little easier if this test of the exposureId was done with the one above.

I.e. do them back to back rather than having them separated by a few tests of exceptions.

tests/test_ts8.py
```diff
@@ -117,6 +116,39 @@ def setUp(self):
...
+        exposureId = self.butler.get('ccdExposureId', dataId={})
+        self.assertEqual(exposureId, 0)
+
+        exposureId = self.butler.get('ccdExposureId', dataId={"visit": 1, "detector": 1})
+        self.assertEqual(exposureId, 101)
+
+        exposureId = self.butler.get('ccdExposureId', dataId={"visit": 1, "detectorName": "S01"})
+        self.assertEqual(exposureId, 101)
+
+        exposureId = self.butler.get('ccdExposureId', dataId={"dateObs": "2018-12-31T05:06:33.54",
```
SimonKrughoff: Same as above.

tests/test_ucdcam.py
```diff
@@ -114,6 +113,41 @@ def setUp(self):
...
+        exposureId = self.butler.get('ccdExposureId', dataId={"visit": 1, "detector": 1})
+        self.assertEqual(exposureId, 11)
+
+        with self.assertRaises(ValueError):
+            self.butler.get('ccdExposureId', dataId={"visit": 1, "detector": 4})
+
+        with self.assertRaises(KeyError):
+            self.butler.get('ccdExposureId', dataId={"visit": 1})
+
+        exposureId = self.butler.get('ccdExposureId', dataId={"visit": 1, "raftName": "R01",
```
SimonKrughoff: And again.

python/lsst/obs/lsst/ingest.py
```diff
@@ -38,15 +38,12 @@ class LsstCamParseTask(ParseTask):
...
 
-    camera = None
     _translatorClass = None
 
     def __init__(self, config, *args, **kwargs):
         super().__init__(config, *args, **kwargs)
 
         self.observationInfo = None
-        if self.camera is None:
-            self.camera = self._mapperClass._makeCamera()
```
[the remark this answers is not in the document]
timj: The mapper makes the camera (that's done in the base class).  This camera was to support the metadata translation.

tests/test_auxTel.py
timj: Why isn't flake8 complaining???

timj: So `import lsstx.daf.base` warns but `import lsst.daf.base` does not. It seems that flake8 is looking at the other `import lsst.` and assuming that they are all related. Weird.

tests/test_parsetask.py
timj: This is a warning from lsst.log so can't be captured until I merge lsst/log#37 (which might happen today: DM-17651).

SimonKrughoff: Fair enough.

timj: Now added the tests.

PR #76, opened 2019-02-06
Fix typo in LsstCamMapper

SimonKrughoff:
This one line change does not allow the default mapper class to be constructed, but it at least removes the file not found error.


SimonKrughoff:
I'm merging.

PR #80, opened 2019-02-13
DM-17875: DESC config updates for DC2 stars only refererence catalog

heather999:
As agreed at this week's DM-DC2 meeting, we want to merge the config changes from the u/jchiang/run2.xi_stars_only_ref_cats branch into master for use as part of the DC2 Run2.1i processing which will use a star-only reference catalog.

Opening the PR, but would like some guidance on what testing is required by the reviewers @RobertLuptonTheGood @timj  I don't think I can start up a Jenkins stack-os-matrix Job - is that something I should request a DM person to do?

@jchiang87 I think I cherry-picked your 2 commits properly, but please take a peek to make sure.

### DIFF
--- a/config/imsim/processCcd.py
+++ b/config/imsim/processCcd.py
@@ -22,7 +22,4 @@
 
 config.isr.doCrosstalk=True
 
-# set SIP order fitting to 2 to avoid bad astrometric solutions
-config.calibrate.astrometry.wcsFitter.order=2
-
 config.calibrate.photoCal.match.referenceSelection.magLimit.fluxField = "i_flux"
--- a/config/imsim/processCcd.py
+++ b/config/imsim/processCcd.py
@@ -21,3 +21,8 @@
 # see <http://www.lsstcorp.org/LegalNotices/>.
 
 config.isr.doCrosstalk=True
+
+# set SIP order fitting to 2 to avoid bad astrometric solutions
+config.calibrate.astrometry.wcsFitter.order=2
+
+config.calibrate.photoCal.match.referenceSelection.magLimit.fluxField = "i_flux"
--- a/config/imsim/processCcd.py
+++ b/config/imsim/processCcd.py
@@ -21,5 +21,3 @@
 # see <http://www.lsstcorp.org/LegalNotices/>.
 
 config.isr.doCrosstalk=True
-
-config.calibrate.photoCal.match.referenceSelection.magLimit.fluxField = "i_flux"
--- a/config/processCcd.py
+++ b/config/processCcd.py
@@ -61,7 +61,6 @@ for refObjLoader in (config.charImage.refObjLoader,
                      config.calibrate.astromRefObjLoader,
                      config.calibrate.photoRefObjLoader):
     refObjLoader.retarget(LoadIndexedReferenceObjectsTask)
-#    refObjLoader.load(os.path.join(getPackageDir('obs_lsst'), 'config', 'filterMap.py'))
     refObjLoader.ref_dataset_name='cal_ref_cat'
 
 # Better astrometry matching
@@ -82,7 +81,6 @@ config.calibrate.photoCal.applyColorTerms = False
 #colors["g-r"] = ColorLimit(primary="g_flux", secondary="r_flux", minimum=0.0)
 #colors["r-i"] = ColorLimit(primary="r_flux", secondary="i_flux", maximum=0.5)
 config.calibrate.photoCal.match.referenceSelection.doMagLimit = True
-#config.calibrate.photoCal.match.referenceSelection.magLimit.fluxField = "lsst_i_smeared_flux"
 config.calibrate.photoCal.match.referenceSelection.magLimit.fluxField = "i_flux"
 config.calibrate.photoCal.match.referenceSelection.magLimit.maximum = 22.0
 # select only stars for photometry calibration
--- a/config/processCcd.py
+++ b/config/processCcd.py
@@ -61,7 +61,7 @@ for refObjLoader in (config.charImage.refObjLoader,
                      config.calibrate.astromRefObjLoader,
                      config.calibrate.photoRefObjLoader):
     refObjLoader.retarget(LoadIndexedReferenceObjectsTask)
-    refObjLoader.load(os.path.join(getPackageDir('obs_lsst'), 'config', 'filterMap.py'))
+#    refObjLoader.load(os.path.join(getPackageDir('obs_lsst'), 'config', 'filterMap.py'))
     refObjLoader.ref_dataset_name='cal_ref_cat'
 
 # Better astrometry matching
@@ -79,7 +79,8 @@ config.calibrate.photoCal.applyColorTerms = False
 #colors["g-r"] = ColorLimit(primary="g_flux", secondary="r_flux", minimum=0.0)
 #colors["r-i"] = ColorLimit(primary="r_flux", secondary="i_flux", maximum=0.5)
 config.calibrate.photoCal.match.referenceSelection.doMagLimit = True
-config.calibrate.photoCal.match.referenceSelection.magLimit.fluxField = "lsst_i_smeared_flux"
+#config.calibrate.photoCal.match.referenceSelection.magLimit.fluxField = "lsst_i_smeared_flux"
+config.calibrate.photoCal.match.referenceSelection.magLimit.fluxField = "i_flux"
 config.calibrate.photoCal.match.referenceSelection.magLimit.maximum = 22.0
 # select only stars for photometry calibration
 config.calibrate.photoCal.match.sourceSelection.unresolved.maximum=0.5
--- a/config/processCcd.py
+++ b/config/processCcd.py
@@ -61,14 +61,12 @@ for refObjLoader in (config.charImage.refObjLoader,
                      config.calibrate.astromRefObjLoader,
                      config.calibrate.photoRefObjLoader):
     refObjLoader.retarget(LoadIndexedReferenceObjectsTask)
+    refObjLoader.load(os.path.join(getPackageDir('obs_lsst'), 'config', 'filterMap.py'))
     refObjLoader.ref_dataset_name='cal_ref_cat'
 
 # Better astrometry matching
 # commented out for now
 #config.calibrate.astrometry.matcher.numBrightStars = 150
-
-# set SIP order fitting to 2 to avoid bad astrometric solutions
-config.calibrate.astrometry.wcsFitter.order=2
 
 # Set to match defaults currenyly used in HSC production runs (e.g. S15B)
 config.charImage.catalogCalculation.plugins['base_ClassificationExtendedness'].fluxRatio = 0.95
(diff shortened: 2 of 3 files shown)


config/processCcd.py
```diff
@@ -61,13 +61,16 @@
                      config.calibrate.astromRefObjLoader,
                      config.calibrate.photoRefObjLoader):
     refObjLoader.retarget(LoadIndexedReferenceObjectsTask)
-    refObjLoader.load(os.path.join(getPackageDir('obs_lsst'), 'config', 'filterMap.py'))
+#    refObjLoader.load(os.path.join(getPackageDir('obs_lsst'), 'config', 'filterMap.py'))
```
timj: Please remove commented out code.

config/processCcd.py
```diff
@@ -79,7 +82,8 @@
 #colors["g-r"] = ColorLimit(primary="g_flux", secondary="r_flux", minimum=0.0)
 #colors["r-i"] = ColorLimit(primary="r_flux", secondary="i_flux", maximum=0.5)
 config.calibrate.photoCal.match.referenceSelection.doMagLimit = True
-config.calibrate.photoCal.match.referenceSelection.magLimit.fluxField = "lsst_i_smeared_flux"
+#config.calibrate.photoCal.match.referenceSelection.magLimit.fluxField = "lsst_i_smeared_flux"
```
timj: Remove.

policy/cameraHeader.yaml
```diff
@@ -76,14 +76,14 @@ AMP_ITL : &AMP_ITL
...
 
     # Linearity correction is still under discussion, so this is a placeholder.
     linearityType : PROPORTIONAL
     linearityThreshold : 0
-    linearityMax : 65535                # == saturation
-    linearityCoeffs : [0, 65535]        # == [linearityThreshold, linearityMax]
-
+    linearityMax : 142857                # == saturation
+    linearityCoeffs : [0, 142857]        # == [linearityThreshold, linearityMax]
+
```
timj: Do not add trailing whitespace.

config/processCcd.py
```diff
@@ -61,13 +61,15 @@
                      config.calibrate.astromRefObjLoader,
                      config.calibrate.photoRefObjLoader):
     refObjLoader.retarget(LoadIndexedReferenceObjectsTask)
-    refObjLoader.load(os.path.join(getPackageDir('obs_lsst'), 'config', 'filterMap.py'))
```
RobertLuptonTheGood: Could you make these changes in `config/imsim/processCcd.py`?  You can load `config/processCcd.py` first and then just make the changes that you want.

In the longer run fixing the top-level config may well be right, but it'll also affect e.g. ts8 and comCam

policy/cameraHeader.yaml
```diff
@@ -54,13 +54,13 @@ AMP_E2V : &AMP_E2V                      # N.b. the assembly of E2V amps into a C
     #v_overscan : 46                     # number of overscan pixel in y (parallel)
     #v_overscan0 : 0                     # first overscan pixel in y (parallel)
 
-    saturation : 56000                  # saturation level, DN XXX Should this be in electrons?
```
RobertLuptonTheGood: This is OK to merge for now, although it affects other cameras.  See DM-17916 for a change to make these numbers per-CCD and per-amp, and DM-17915 for converting these values to be in electrons, not DN.  Neither of these will make it unto 17.0 (although I could get them in soon if desired by DESC)

jchiang87: > Neither of these will make it unto 17.0 (although I could get them in soon if desired by DESC)

Thanks.  This is sufficient for now.  Eventually, imSim will get these numbers from `obs_lsst`, especially if they are made to be per-CCD or per-amp, but we're not there yet.

config/imsim/processCcd.py
```diff
@@ -21,3 +21,8 @@
 # see <http://www.lsstcorp.org/LegalNotices/>.
 
 config.isr.doCrosstalk=True
+
+# set SIP order fitting to 2 to avoid bad astrometric solutions
```
heather999: I've moved these 2 new configs into imsim/processCcd.py

config/processCcd.py
heather999: I've reintroduced this line (and move the other 2 new configs into imsim/processCcd.py).  How do I effectively undo this loading of filterMap.py for the imSim case @RobertLuptonTheGood  @jchiang87 ?  is there such a thing as refObjLoader.unload or do we just have to reset config.filterMap to some default within imsim/procesCcd?

RobertLuptonTheGood:
Good question.  I don't know, and if this is hard I withdraw my suggestion for now.

The filters should be being loaded in the per-camera configs.

heather999:
It might not be hard - but I need some DM config expert to provide some guidance. If we're all in agreement to move the filter loading to the per-camera configs then it might be easy.  Poking @timj

timj:
I'm not a config expert. This was all new to me a couple of days ago. Maybe @ktlim or @TallJimbo ?

config/imsim/processCcd.py
jchiang87: I thought the consensus was that setting the `wcsFitter.order=2` is something we wanted to do for all (LSST) data, not just for imSim data.  Maybe this needs to be fixed outside this package, but I'm not sure just putting it here is what we want.  At the very least, if it lives here, it should also be in `phosim/processCcd.py`.

TallJimbo: Yes, this should go in just `config/processCcd.py`.  It should be included for all LSST data for which fitting a WCS is meaningful, and if it's there it will be ignored by things like TS8 for which WCS-fitting never happens.

config/processCcd.py
TallJimbo: You should be able to reset a `Config` to its defaults by setting it to a default-constructed instance, e.g.
```
config.calibrate.refObjLoader = LoadIndexedReferenceObjectsConfig()
```
That said, updating and then resetting a config to its defaults is not a pattern I'd recommend; if the goal is to set something for just one camera, it definitely would be better to put it in the override directory for just that camera.  If it's relevant for a subset of the cameras (say, `imsim` and `phosim`, but not the test stands or AuxTel), I'd gather those options into a new file (name it whatever you want) in `config/` and load it explicitly in e.g. the files in both `config/imsim/` and `config/phosim/`

config/imsim/processCcd.py
heather999: Ok - done

config/processCcd.py
jchiang87: I was just chatting with @erykoff about https://jira.lsstcorp.org/browse/DM-17917.  If that work is included in v17, then I don't think we need this particular change since it's meant to support using a stars-only reference catalog.  Plus, I think we could also omit the change to 
```
config.calibrate.photoCal.match.referenceSelection.magLimit.fluxField = "i_flux"
```
In summary, the only one we would need in this PR is the `wcsFitter.order=2` change.

There would, of course, be additional config lines we'd need to use to take advantage of DM-17917, but IMO, those don't necessarily need to be the defaults in this package.

timj: DM-17917 would have to be reviewed and merged today for it to be in the v17.0 release.

heather999: Unfortunately, I'm not sure what setting of filterMap is needed for which camera.  It looks to me that `imsim` and `phosim` want `filterMap` to take the global default (is that right @jchiang87 ?). Meanwhile ts8? ts3? (maybe others?) want to set filterMap according to the contents of `obs_lsst/config/filterMap.py`: `config.filterMap = {band: 'lsst_%s_smeared' % (band) for band in 'ugrizy'}`.  So I think the way forward is to load `obs_lsst/config/filterMap.py` inside `obs_lsst/config/ts8/processCcd.py` and `obs_lsst/config/ts3/processCcd.py` or whatever cameras need it.

erykoff: @timj That's why I'm begging for a reviewer, it's only a few lines!

timj:
@heather999 the config hierarchy is less than optimal at the moment and has grown organically. It definitely needs a clean sweep. The only criteria for "working" we have at the moment is whether ci_lsst runs okay but ci_lsst does not yet include full sets of calibration data for all supported cameras. I think that for now having phosim/ and imsim/processCcd.py is the best short term compromise if you want something in today.

config/processCcd.py
heather999: Ok - catching up with the comments.. I'll hold off until I see what happens with DM-17917.

jchiang87: `filterMap.py` would only be relevant for sky exposures, so ts8 and ts3 won't need it.

timj: Whilst you are waiting (for something that might not happen) please see if you can get ci_lsst working now.

heather999: I'll retry ci_lsst.
@RobertLuptonTheGood which cameras need `filterMap.py`?

erykoff: DM-17917 is merged.  The usage relevant for DC2 is then of this form:
```
config.calibrate.astrometry.referenceSelection.doUnresolved = True
config.calibrate.astrometry.referenceSelection.unresolved.name = 'unresolved_name'
config.calibrate.astrometry.referenceSelection.unresolved.minimum = minimum_val
config.calibrate.astrometry.referenceSelection.unresolved.maximum = maximum_val
```
And then only objects with `unresolved_name` (I don't know the field name in the reference catalogs) between `minimum_val` and `maximum_val` will be selected for astrometric matching.

heather999: So going back to @jchiang87's comment.. 
I will reintroduce `refObjLoader.load(os.path.join(getPackageDir('obs_lsst'), 'config', 'filterMap.py'))` in `obs_lsst/config/processCcd.py` and remove `config.calibrate.photoCal.match.referenceSelection.magLimit.fluxField = "i_flux"` from `obs_lsst/config/imsim/processCcd.py`  Then I'll try `ci_lsst` again, and we might be good to go :)

RobertLuptonTheGood: All cameras do, but your filter names aren't useful for non-imsim/phosim cameras (e.g. comCam will have LSST cameras;  auxTel is different!).  So please move the filterMap.py to a place that you can use it, and don't worry about other cameras

heather999: The `ci_lsst` run completed successfully as far as I can see.  The output log is here if anyone is interested: `/home/heatherk/work/ci_lsst/running_imsim_20190215.log`

This PR has been modified such that the only change to `config/processCcd.py` is the addition of the agreed to line: `config.calibrate.astrometry.wcsFitter.order=2` and the changes to `policy/cameraHeader.yaml`  The change to remove loading `config/filterMap.py` has been removed and the handling for `config/filterMap.py is identical to what is on master. Is this possibly good to go @jchiang87 & @RobertLuptonTheGood?

heather999:
The `ci_lsst` run completed successfully as far as I can see. There are no error messages in the log. The output log is here if anyone is interested: `/home/heatherk/work/ci_lsst/running_imsim_20190215.log`

This PR has been modified such that the only change to `config/processCcd.py` is the addition of the agreed to line: `config.calibrate.astrometry.wcsFitter.order=2` and the changes to `policy/cameraHeader.yaml` The change to remove loading `config/filterMap.py` has been removed and the handling for `config/filterMap.py` is identical to what is on master. Is this possibly good to go @jchiang87 & @RobertLuptonTheGood?

PR #81, opened 2019-02-14
DM-17794: Update tests and visit ID calculations

timj:
* Shrink some of the exposure_id and detector_exposure_id calculations
* Add missing TS3 test
* Check that the detector_exposure_id can fit in allocated number of bits (enlarging some values).


python/lsst/obs/lsst/lsstCamMapper.py
```diff
@@ -365,7 +365,7 @@ def bypass_ccdExposureId(self, datasetType, pythonType, location, dataId):
 
     def bypass_ccdExposureId_bits(self, datasetType, pythonType, location, dataId):
         """How many bits are required for the maximum exposure ID"""
-        return 36  # just a guess, but this leaves plenty of space for sources
+        return 51  # max detector_exposure_id ~ 2050121299999250
```
SimonKrughoff: Why use 51 here when we are using 58 everywhere else?  If we use 58 here, can't we just use this function in all the mappers by inheritance?

python/lsst/obs/lsst/translators/lsstsim.py
```diff
@@ -58,7 +58,7 @@ def compute_detector_exposure_id(exposure_id, detector_num):
         detector_exposure_id : `int`
             The calculated ID.
         """
-        return compute_detector_exposure_id_generic(exposure_id, detector_num, max_num=1000,
+        return compute_detector_exposure_id_generic(exposure_id, detector_num, max_num=999,
```
SimonKrughoff: This change seems a bit strange.  The default is 1000.  I'd understand if the edit was to change to use the default, but one less than the default seems odd.

python/lsst/obs/lsst/ucd.py
```diff
@@ -53,6 +53,10 @@ def _extractDetectorName(self, dataId):
             raftName = LsstUCDCamTranslator.compute_detector_group_from_num(dataId['detector'])
         return f"{raftName}_{detectorName}"
 
+    def bypass_ccdExposureId_bits(self, datasetType, pythonType, location, dataId):
+        """How many bits are required for the maximum exposure ID"""
+        return 48  # max detector_exposure_id ~ 203012122359590
```
SimonKrughoff: I find the comment here a little odd.  It has a tilde, implying of order, but then has a very specific number.  Additionally, I'd assume that number would just be `2**48`, but it's not.  The other cases are like this too, and I find them equally odd.

python/lsst/obs/lsst/translators/auxTel.py
```diff
@@ -180,7 +180,8 @@ def compute_exposure_id(dayobs, seqnum):
         dayobs = dayobs.replace("-", "")
 
         # Form the number as a string zero padding the sequence number
-        idstr = f"{dayobs}{seqnum:06d}"
+        # Expect no more than 99,999 exposures in a day
```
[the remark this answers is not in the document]
timj: This is auxTel and lsstCam. We won't be taking more than one per second (86400 images). I believe the guider takes a data cube of images rather than writing a file per image (but we can always use different visit definition for guider data).

python/lsst/obs/lsst/translators/lsstsim.py
timj: Because in concat mode it reserves enough space to represent the maximum number. Going from 1000 to 999 is the same as going from 1000 to ~ 230. People keep changing their mind as to the most number of detectors we have to support and using 999 covers all options.

python/lsst/obs/lsst/ucd.py
timj: I am saying that I wouldn't expect a detector exposure ID for UCDCam data after 2030-12-12 but it's a ~ because it's a guess. I could just as easily say 2022-12-12 I would think but test stand data doesn't need many sources for a source ID.

python/lsst/obs/lsst/lsstCamMapper.py
timj: Why would I waste bits when I don't need to? Phosim/AuxTel/LSSTCam/Imsim all are meant to be using the same definition of exposure ID. The test stands use a timestamp and have much higher bit requirements. The community post talks about this and makes clear that even 51 is too high the way things are currently set up for source ID generation but I think it's important to note that the exposure IDs with the different cameras are inherently different.

python/lsst/obs/lsst/ucd.py
SimonKrughoff: OK.  I didn't realize there was a date in there.  Thanks for the explanation.

python/lsst/obs/lsst/translators/auxTel.py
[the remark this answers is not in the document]
timj: I just pushed a change to check that the sequence number really is less than 99,999 so that we can complain if something like guider data does turn up.

PR #82, opened 2019-02-14
DM-17903: Ensure that ccdKeys is set

timj:
#73 wasn't quite correct. It ended up setting the ccdKeys in config.isr rather than the root config. This broke imsim ci_lsst.  With this change I reintroduce a generic lsstCamCommon configuration that defines ccdKeys.

### DIFF
--- a/config/dark.py
+++ b/config/dark.py
@@ -23,6 +23,7 @@
 import os.path
 from lsst.utils import getPackageDir
 
+config.load(os.path.join(getPackageDir("obs_lsst"), "config", "lsstCamCommon.py"))
 config.isr.load(os.path.join(getPackageDir("obs_lsst"), "config", "isr.py"))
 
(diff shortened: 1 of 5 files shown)


config/dark.py
```diff
@@ -23,6 +23,7 @@
 import os.path
 from lsst.utils import getPackageDir
 
+config.load(os.path.join(getPackageDir("obs_lsst"), "config", "lsstCamCommon.py"))
```
RobertLuptonTheGood: Is there a reason not to load it from `isr.py`?

timj: Before this patch the problem was exactly that we were setting this from isr.py. Wouldn't we have the same problem if we did that. My understanding is that the problem is that we were setting it in isr.py and it was setting the value in `config.isr` not `config`.

PR #86, opened 2019-03-08
DM-18329: Add jointcal wcs and photoCalib templates.

erykoff:


policy/lsstCamMapper.yaml
```diff
@@ -332,6 +332,10 @@ datasets:
     storage: FitsCatalogStorage
     tables: raw
     template: srcMatch/%(visit)08d-%(filter)s/%(raftName)s/srcMatch_%(visit)08d-%(filter)s-%(raftName)s-%(detectorName)s-det%(detector)03d.fits
+  jointcal_wcs:  # astrometric calibration produced by jointcal
+    template: jointcal-results/%(filter)s/%(tract)04d/jointcal_wcs-%(visit)07d-%(raftName)s-%(detectorName)s-det%(detector)03d.fits
+  jointcal_photoCalib:  # photometric calibration produced by jointcal
+    template: jointcal-results/%(filter)s/%(tract)04d/jointcal_photoCalib-%(visit)07d-%(raftName)s-%(detectorName)s-det%(detector)03d.fits
```
parejkoj: Do we really need `raftName`, `detectorName` and `detector` in here? Wouldn't either the first two, or just the last one, be sufficient?

erykoff: Good question.  Maybe @jchiang87 can chime in because he wrote these, but if you look at the rest of the policy file this seems to be the standard template, including the raw exposure template.  So I think it should be left as-is, and whether to simplify this would have to be decided at another level.

jchiang87: I did try just specifying the first two, and I recall it failing for me, though I don't remember the exact issue.  In any case, as Eli notes, I just followed the convention for the other templates and everything worked fine.

PR #87, opened 2019-03-19
DM-17866: Add support for BOT/lsstCam

timj:


python/lsst/obs/lsst/translators/lsstCam.py
```diff
@@ -0,0 +1,86 @@
...
+    # Use Imsim raft definitions until a true lsstCam definition exists
+    cameraPolicyFile = "policy/lsstCam.yaml"
+
+    @cache_translation
+    def to_science_program(self):
+        # Docstring will be inherited. Property defined in properties.py
+        # Do not yet have a header explicitly for this.
+        program = None
+        if "TSTAND" in self._header and "DAYOBS" in self._header:
+            program = self._header["DAYOBS"]
```
RobertLuptonTheGood: Can we set program to `None` or `"unknown"` or something rather than `DAYOBS`?  I don't think that `program` has any LSST meaning, and we shouldn't be relying on it.

tests/test_translator.py
```diff
@@ -34,6 +34,54 @@
...
+                           pressure=None,
+                           relative_humidity=None,
+                           science_program=None,
+                           temperature=None,
+                           visit_id=2019031900001)),
+                     )
+        for filename, expected in test_data:
+            with self.subTest(f"Testing {filename}"):
+                self.assertObservationInfoFromYaml(filename, dir=self.datadir, **expected)
+
```
RobertLuptonTheGood: What is this testing?  On the face if it, that we can read ordered dicts from yaml but there must be something more.  A doc string would help

python/lsst/obs/lsst/translators/lsstCam.py
timj: This gets converted to gen2 `run` concept and ingest fails if `run` is None so I have to pick something.  LSST does have a science program meaning since we will be distinguishing between all the deep drilling fields and fast/wide/deep.  This definition of `run` is consistent with what is happening in HSC translation.

tests/test_translator.py
timj: This is standard boiler plate for astro_metadata_translator testing.  The raw headers are read in and translated and the results of the translation are compared with the supplied dict.

python/lsst/obs/lsst/translators/lsstCam.py
timj: I'll switch to "unknown" (we used dayObs for UCD translator).

PR #91, opened 2019-03-28
DM-18051: Get defect machinery working for the AuxTel

mfisherlevine:


bin.src/genDefectFits.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def genDefectFits(cameraName, source, targetDir):
+    mapper = mapperMap[cameraName](root=".", calibRoot=".")
+    camera = mapper.camera
+
+    ccds = dict()
+    for ccd in camera:
+        ccdNum = ccd.getId()
+        ccdName = mapper._extractDetectorName(None)
```
SimonKrughoff: I'm uncomfortable getting the ccd name two different ways.  I.e. why is this not `ccdName = ccd.getName()`?

bin.src/genDefectRegistry.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python
+
+import glob
+import os
+import re
+try:
```
SimonKrughoff: I don't understand why this protection is necessary.  `sqlite3` is the name of the module on both my mac and a linux instance I have.

config/auxTel/flat.py
```diff
@@ -0,0 +1,26 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+
+import os.path
+from lsst.utils import getPackageDir
+
+config.load(os.path.join(getPackageDir("obs_lsst"), "config", "auxTel", "auxTel.py"))
```
SimonKrughoff: This and the following config seem a little backward to me.  It seems like task specific info should be in these files and then aggregated into a larger config file, if necessary.

python/lsst/obs/lsst/lsstCamMapper.py
```diff
@@ -339,6 +339,14 @@ def _extractDetectorName(self, dataId):
 
         return "%s_%s" % (raftName, detectorName)
 
+    def _defectLookup(self, dataId):
+        """Find the defects for a given CCD.
+
+        This method just exists to supplement the dateKey, as we don't have
+        the default, taiObs, in the obs_lsst registry.
+        """
+        return super()._defectLookup(dataId, dateKey='dayObs')
```
SimonKrughoff: I don't understand how this can work since the function in the base class does not take `**kwargs`.

SimonKrughoff: Never mind.  This is added in the other PR for this ticket.  I just hadn't reviewed that yet.

timj:
Am I right in thinking that a bunch of this code is copied directly from obs_subaru? At first glance that doesn't seem like a sustainable approach to development so why are we copying code around rather than putting it in obs_base or somewhere?

config/auxTel/dark.py
```diff
@@ -0,0 +1,33 @@
...
+# see <http://www.lsstcorp.org/LegalNotices/>.
+
+import os.path
+from lsst.utils import getPackageDir
+
+config.load(os.path.join(getPackageDir("obs_lsst"), "config", "auxTel", "auxTel.py"))
+
+<<<<<<< HEAD
+config.repair.cosmicray.nCrPixelMax = 500000
+=======
```
timj: This merge conflict needs to be sorted out.

timj:
Please remove the bad merge commit on this branch.

config/auxTel/dark.py
mfisherlevine: All the code added by Andrés needs to be sorted out actually - it was supposed to be on DM-18749 not here.

mfisherlevine:
> Am I right in thinking that a bunch of this code is copied directly from obs_subaru? At first glance that doesn't seem like a sustainable approach to development so why are we copying code around rather than putting it in obs_base or somewhere?

Yes, that's correct, both that it was copied, and that it's not sustainable. However, it's a pattern that shouldn't be followed in general, so was decided that it wouldn't be moved to `obs_base` precisely for that reason, _i.e._ to discourage others from doing this. Once we live in the glorious world of Gen3 and post-calibration-product-production-overhaul things like this will be cleaned up/stop happening.

config/auxTel/flat.py
mfisherlevine: I'm not sure I've totally understood what you mean, but I did just make a change here, and I think it _might_ address this.

All the config options which are the same for the calib construction tasks now lives in `config/auxTel.py`, and the only one which currently makes any tweaks is `dark.py` as we expect more cosmics in darks.

Does this seem OK now?

bin.src/genDefectFits.py
mfisherlevine: I imagine that in some other cameras there is some terrifying/disgusting reason for this, and that this was just cargo-culted in (as it was all basically just copypasta). Have changed this now (and the following line too, as that was equally dumb once this was changed).

bin.src/genDefectRegistry.py
mfisherlevine: Right, but what if it _wasn't_?

Also, I have no idea really, I just copied and pasted it, and I have no real idea of the differences between sqlite and sqlite3. Perhaps this was coded in an age where the two frequently coexisted, and now the old one never does?

Either way though, it seems pretty harmless to have a fallback to a different version in some cases. Do you think this should be removed?

SimonKrughoff: I'm not overly hung up about it.

config/auxTel/flat.py
SimonKrughoff: I was just saying it seemed like there was configuration just for darks in `auxTel.py`, but I think I understand now that there are config options that you use in lots of tasks and only override task specific things in the task specific config.
I'm fine with that.

PR #92, opened 2019-03-29
DM-18708: Add support for BOT RUNNUM and FILTER2 headers

timj:


python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -406,3 +406,26 @@ def to_exposure_id(self):
...
+        """Calculate the physical filter name.
+
+        Returns
+        -------
+        filter : `str`
+            Name of filter. Can be a combination of FILTER, FILTER1 and FILTER2
+            headers.  Returns "NONE" if no filter is declared.
+        """
+        filterNames = []
+        for k in ("FILTER", "FILTER1", "FILTER2"):
```
SimonKrughoff: Is there any way to make this configurable so you don't have to edit code to add a filter?

timj: astro_metadata_translator translators should not depend on pex_config. Also, obs_lsst cameras are not meant to be using different filter keywords at every opportunity.

SimonKrughoff: I didn't necessarily mean using `pex_config`.  You can do what you want, I just noticed that was one of the changes in this change set (adding a filter).

PR #93, opened 2019-04-03
DM-18748: Support new AT files

timj:
That have missing EXPTIME and GROUPID meaning IMGTYPE.


python/lsst/obs/lsst/translators/auxTel.py
```diff
@@ -169,6 +167,34 @@ def to_location(self):
             return AUXTEL_LOCATION
         return None
 
+    @cache_translation
+    def to_dark_time(self):
+        # Docstring will be inherited. Property defined in properties.py
+        if self.is_key_ok("DARKTIME"):
+            return self.quantity_from_card("DARKTIME", u.s)
+
+        return self.to_exposure_time()
```
mfisherlevine: Shouldn't there be some sort of warning if the `darkTime` is inferred from the expTime?

python/lsst/obs/lsst/translators/auxTel.py
```diff
@@ -169,6 +167,34 @@ def to_location(self):
...
+        # Some data is missing a value for EXPTIME.
+        # Have to be careful we do not have circular logic when trying to
+        # guess
+        if self.is_key_ok("EXPTIME"):
+            return self.quantity_from_card("EXPTIME", u.s)
+
+        # Guess from start end times (and hope that end time is not derived
+        # from exposure time)
+        try:
+            delta = self.to_datetime_end() - self.to_datetime_begin()
```
mfisherlevine: I think that this is often going to be highly suspect/inaccurate, and it worries me deeply. It's not wrong in the code sense, but I really don't like it.

python/lsst/obs/lsst/translators/auxTel.py
```diff
@@ -169,6 +167,34 @@ def to_location(self):
...
+        except RecursionError:
+            log.warning("Insufficient information to derive exposure time. Setting to 0.0s")
+            return 0.0 * u.s
```
mfisherlevine: I think that defaulting to zero instead of this being a fatal error is a huge mistake, especially as we don't have good info telling us which are biases and which are not. I don't think this is OK to merge.

python/lsst/obs/lsst/translators/auxTel.py
```diff
@@ -181,10 +207,20 @@ def to_observation_type(self):
...
+
+        # AuxTel observation type is documented to appear in OBSTYPE
+        # but for historical reasons prefers IMGTYPE.  Some data puts
+        # it in GROUPID (which is meant to be for something else).
+        test_keys = ["OBSTYPE", "IMGTYPE"]
+
+        if not self._is_on_mountain():
+            test_keys.append("GROUPID")
+
+        for k in test_keys:
```
mfisherlevine: This is fragile, as it makes the order that the list is in potentially important, _i.e._ if there is more than one key present. I think that's probably acceptable, but can you add a comment that the order of that list is the order of precedence, or even better, if you can think of a good name, bake it into the variable name?

PR #94, opened 2019-04-04
DM-17830: Use only stars for jointcal internal fitting

yalsayyad:


yalsayyad:
The configs originally on this PR are being made default for all cameras.

PR #95, opened 2019-04-04
DM-18746: Remove list of CALIB tables from ingestCalib.py and use the default to avoid omitting sky table

heather999:
It was discovered that we are overriding the default list of CALIB tables and have omitted the sky table.  To avoid this in the future, we will remove this one line from `config/ingestCalib.py`, as discussed on the dm-lsstcam channel: https://lsstc.slack.com/archives/CBE964PR8/p1554124198054300


heather999:
@RobertLuptonTheGood I just rebased this PR in the hopes that will help get this merged :)

heather999:
This issue will be handled in DM-19270 and its associated PR.

PR #101, opened 2019-04-16
Added new raft defintions and mappings for TS8 data-taking

eacharles:
There the raft slot definitions that I extracted from the eTraveler.   I tested RTM-023 with some recent TS8 data and it works fine.   We don't have TS8 data for the other rafts yet.

### DIFF
--- a/policy/ts8/RTM-013.yaml
+++ b/policy/ts8/RTM-013.yaml
@@ -1,0 +1,186 @@
+RTM-013 :
+  detectorType : ITL
+  raftSerial : LCA-11021_RTM-013
+  ccdSerials :
+    S00 : ITL-3800C-269
+    S01 : ITL-3800C-261
+    S02 : ITL-3800C-205
+    S10 : ITL-3800C-160
+    S11 : ITL-3800C-244
+    S12 : ITL-3800C-157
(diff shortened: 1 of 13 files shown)


timj:
Are you deliberately using a bad email address for your git commit?

eacharles:
No it certainly isn't deliberate.   What is the email address in question?

timj:
See https://github.com/lsst/obs_lsst/commit/ba27404ea2109f439e816da150bc00892a247ded.patch because it looks like you are `badass@stanford.edu` and maybe that is fine??? (GitHub doesn't recognize you).

eacharles:
Ah, yes, that is in fact my stanford email address.   For github I use my gmail one.   I'll change the .gitconfig on that machine to avoid confusion.

timj:
Jenkins tests are failing because one of the tests asks how many detectors there are in ts8 (test_ts8.py). That needs to change from 98 to 198.

eacharles:
Ok, I did fixed the ndetectors in tests/test_ts8.py.   Where are the jenkins test results?   I only see a couple of travis-ci tests in the check section above and those appear to be pretty minimal.

PR #105, opened 2019-04-25
DM-16297: Add Gen3 butler support

timj:


python/lsst/obs/lsst/assembly.py
```diff
@@ -0,0 +1,203 @@
...
+from lsst.afw.geom import makeSkyWcs
+from lsst.ip.isr import AssembleCcdTask
+import lsst.afw.cameraGeom as cameraGeom
+
+logger = lsst.log.Log.getLogger("LsstCamAssembler")
+
+
+def makeWcsFromDetector(detector, boresight, rotation=0*degrees, flipX=False):
+    """Given a camera, detector and (boresight, rotation), return that
+    detector's WCS
```
parejkoj: This functionality is going to live in `obs_base` as part of DM-20154, which is currently in review: https://github.com/lsst/obs_base/pull/156

timj:
This PR requires lsst/obs_base#170 and lsst/daf_butler#173

python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -138,6 +138,10 @@ class LsstBaseTranslator(FitsTranslator):
     detectorSerials = None
     """Mapping of detector serial number to raft, number, and name."""
 
+    DETECTOR_MAX = 999
```
czwa: This is actually a maximum detector index to use, correct?  Having this be 0 for latiss confused me due to this.

timj: Yes. It's the maximum detector ID we can use before we change the size of the calculated detector_exposure_id

PR #106, opened 2019-05-01
DM-19570: Skip the raw_sub test from obs_base

timj:
raw_sub does not work with obs_lsst package.


timj:
Gen2 is now deprecated so closing this PR.

PR #119, opened 2019-07-06
DM-20069: Add config and dataset for PTC task and metadata

mfisherlevine:


config/measurePhotonTransferCurve.py
```diff
@@ -0,0 +1,20 @@
...
+
+config.load(os.path.join(getPackageDir("obs_lsst"), "config", "lsstCamCommon.py"))
+config.isr.load(os.path.join(getPackageDir("obs_lsst"), "config", "isr.py"))
+
+config.ccdKey = 'detector'
+
+config.isr.doFlat = False
+config.isr.doFringe = False
+config.isr.doCrosstalk = False
+config.isr.doAddDistortionModel = False
```
mfisherlevine: Everything from this point downwards should live in the task's `setDefaults()` option I think, and especially the saturation stuff.

mfisherlevine: Actually, I think all of the isr stuff should actually.

czwa: That sounds reasonable to me as well.

config/findDefects.py
```diff
@@ -0,0 +1,23 @@
+import os.path
```
czwa: Is this an orphan from another ticket?  This (and the template below) do not seem associated with the PTC work.

policy/lsstCamMapper.yaml
```diff
@@ -463,3 +467,11 @@ datasets:
     storage: FitsStorage
     tables: raw
     template: binned_sensor/%(dstype)s-v%(visit)08d/%(raftName)s/%(detectorName)s_%(half)s.fits
+  newDefects:
```
czwa: This template.

PR #120, opened 2019-07-23
tickets/DM-19709 Add config and dataset for defect finding task

mfisherlevine:


config/findDefects.py
```diff
@@ -0,0 +1,18 @@
+import os.path
+from lsst.utils import getPackageDir
+
+config.load(os.path.join(getPackageDir("obs_lsst"), "config", "lsstCamCommon.py"))
+config.isrForFlats.load(os.path.join(getPackageDir("obs_lsst"), "config", "isr.py"))
+config.isrForDarks.load(os.path.join(getPackageDir("obs_lsst"), "config", "isr.py"))
+
+config.ccdKey = 'detector'
+config.assertSameRun = True
```
parejkoj: A comment for why this config is set, please?

config/findDefects.py
```diff
@@ -0,0 +1,18 @@
...
+
+for config in [config.isrForDarks, config.isrForFlats]:
```
parejkoj: Please add a comment about why these are disabled here (not for biases?).

mfisherlevine: I have commented on why in the code. The "not for biases" part doesn't make sense though - this is the two custom isr tasks for processing the darks and flats to find defects in, so there isn't one for biases.

PR #125, opened 2019-09-05
DM-21186: Rename auxTel to LATISS

timj:
The telescope is AuxTel but the camera on that telscope is
called LATISS.


python/lsst/obs/lsst/translators/latiss.py
```diff
@@ -32,7 +32,7 @@
 # Date instrument is taking data at telescope
 # Prior to this date many parameters are automatically nulled out
 # since the headers have not historically been reliable
-TSTART = Time("2019-10-01T00:00", format="isot", scale="utc")
+TSTART = Time("2020-01-01T00:00", format="isot", scale="utc")
```
mfisherlevine: It's quite possible that we'll have data from the mountain before this new date you've selected. I imagine there was a reason you didn't like it being set to this October, but I think pushing it past Nov 1st is dangerous. Could you bring this back a bit please?

PR #126, opened 2019-09-09
DM-21187: Reorganize bin scripts so they can be tested

timj:


python/lsst/obs/lsst/script/phosimToRafts.py
```diff
@@ -0,0 +1,209 @@
...
+
+        if raftName in ("R00", "R40", "R04", "R44"):
+            continue
+        if outputDir:
+            prefix = outputDir
+        else:
+            prefix = os.path.curdir
+        outfile = os.path.join(prefix, f"{raftName}.yaml")
+        logger.debug("Writing raft information to %s", outfile)
+        with open(os.path.join(prefix, f"{raftName}.yaml"), "w") as fd:
```
czwa: Any reason not to use `outfile` here?

timj:
Sorry. I should have noted that most of the changes are moving code around and not new code at all.

python/lsst/obs/lsst/script/phosimToRafts.py
timj: No. I didn't look at that code but it does indeed seem odd to not use outfile.

PR #131, opened 2019-09-19
Work for Aaron

RobertLuptonTheGood:
This is the work that I did for Aaron in response to some queries about the original code.  As you'll see from the commit message it's not really ready to be merged, but more importantly I'm waiting for someone from the camera to confirm that this resolves their problems,


timj:
Who do we need to look into this? I think this might need some work to bring up to date.

PR #123, opened 2019-08-29
Pushing minor changes to add new E2V CCD

craiglagegit:
We have added a new E2V CCD to the UCDavis camera.  This PR adds the support for this new CCD.

PR #133, opened 2019-09-19
DM-21365: Fix test breakage

mfisherlevine:


timj:
This fixes a problem introduced in #123

PR #134, opened 2019-10-07
Merge dc2/run2.1 DESC updates into new dc2/run2.2 branch

heather999:
For DC2 Run2.2i, DESC is moving to use DM 18.1.0.  Merging in updates from our dc2/run2.1 branch into the new dc2/run2.2 branch which is based on 18.1.0

### DIFF
--- a/config/isr.py
+++ b/config/isr.py
@@ -30,3 +30,4 @@ config.doCrosstalk=True
 config.doAddDistortionModel = False
 config.qa.doThumbnailOss = False
 config.qa.doThumbnailFlattened = False
+config.doCrosstalkBeforeAssemble = False
--- a/config/imsim/coaddDriver.py
+++ b/config/imsim/coaddDriver.py
@@ -1,0 +1,23 @@
+# This file is part of obs_lsst.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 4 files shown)


johannct:
@jchiang87  not yet. I will take care of it as soon as I am done with the dr1b metacal rerun request

config/isr.py
```diff
@@ -30,3 +30,4 @@
 config.doAddDistortionModel = False
 config.qa.doThumbnailOss = False
 config.qa.doThumbnailFlattened = False
+config.doCrosstalkBeforeAssemble = False
```
mfisherlevine: This is reasserting the upstream default, which is usually a no-no. This is not proposed as a merge back to the main DM codebase, nor will it be, right? Just wanted to check in on this line as a word of caution, and to check what the final resting place for this line is intended to be.

mfisherlevine: (Had this only been to the imsim isr config I wouldn't have said anything, fwiw)

heather999: This is only for the dc2/run2.2 branch.  Any changes that we intend to merge back to master will likely happen after Gen-3 is more fully ready and will be coordinated with the DM team.

We never worked out getting the config.doCrosstalkBeforeAssemble setting to stick, unless we set it in the upper level config/isr.py.  I am hoping this will become a non-issue with upcoming versions of DM.

mfisherlevine: OK, not a problem then, though that is interesting/worrying about not being able to get it to stick 🤔
I imagine that when any significant merge-back happens I'll be paying attention then, so will file under "ignore for now" 🙂

heather999:
@johannct Would you prefer that I hold off on a merge until you test?  I suspect it might be easier to test, if the changes are available in a dc2/run2.2 branch?

As for DIA template updates, I'm keeping an eye on https://github.com/lsst/obs_lsst/pull/122 The hope is to see that merged into obs_lsst master, and then we use that as our test case to work out backporting a simple update back into the 18.1.0 release.  So, I'm not pulling that into this branch right now.

johannct:
@heather999 you can go ahead with the merge. We will commit to it if needs be

PR #135, opened 2019-10-07
DM-19903: Add columnar defect identification to findDefectsTask in cp_pipe please

plazas:
Added correction files to "corrections" directory with missing keywords for auxTel data from 09/27/19


timj:
The hope is that we improve with time...

PR #143, opened 2019-11-08
DM-21981: Update rawFormatter to correctly resize amplifier bboxes.

czwa:


python/lsst/obs/lsst/gen3/rawFormatter.py
```diff
@@ -77,11 +77,41 @@ def readImage(self):
         image : `~lsst.afw.image.Image`
             In-memory image component.
         """
+        return self.readFull().getImage()
+
+    def readFull(self, parameters=None):
```
timj: Should probably document that there are no acceptable parameters for this readFull implementation.

python/lsst/obs/lsst/gen3/rawFormatter.py
```diff
@@ -77,11 +77,41 @@ def readImage(self):
...
-        return exposure.getImage()
+
+        mask = self.readMask()
+        if mask is not None:
+            exposure.setMask(mask)
+        variance = self.readVariance()
+        if variance is not None:
+            exposure.setVariance(variance)
+
+        info = exposure.getInfo()
```
timj: Which of these lines fixed the real problem with amplifiers?

tests/test_assemble.py
```diff
@@ -57,6 +57,34 @@ def assertAmpRawBBoxesEqual(self, amp1, amp2):
         self.assertEqual(amp1.getRawHorizontalOverscanBBox(), amp2.getRawHorizontalOverscanBBox())
         self.assertEqual(amp1.getRawVerticalOverscanBBox(), amp2.getRawVerticalOverscanBBox())
 
+    def assertAmpRawBBoxesFlippablyEqual(self, amp1, amp2):
```
timj: Please add a doc string.

python/lsst/obs/lsst/gen3/rawFormatter.py
czwa: The fix is in not dropping the exposure; the code previously called readImage(), which did all that is now in readFull(), but returned exposure.getImage().  This dropped the updated amplifier information constructed in fixAmpsAndAssemble(), resulting in an image that contained all of the amplifier image data, but without the updated bbox information to find it.

PR #145, opened 2019-11-13
DM-21169: Add corner rafts

SimonKrughoff:


python/lsst/obs/lsst/script/generateCamera.py
```diff
@@ -78,6 +79,30 @@ def build_argparser():
     return parser
 
 
+def applyRaftYaw(offset, raftYaw):
+    """Apply raft yaw angle to internal offsets of the CCDs.
+
+    Parameters
+    ----------
+    ccdLayout : `list`
```
timj: This parameter has the wrong name.

python/lsst/obs/lsst/script/generateCamera.py
```diff
@@ -78,6 +79,30 @@ def build_argparser():
...
+        A list of the offsets to rotate: [x, y]
+    raftYaw : `float`
+        Raft yaw angle in degrees.
+
+    Returns
+    -------
+    2-item sequence of floats containing the rotated offsets.
```
timj: Return should explicitly show each item separately.

python/lsst/obs/lsst/script/generateCamera.py
```diff
@@ -193,18 +218,22 @@ def indent():
...
                 if ccdName in geometryWithinRaft:
                     doffset = geometryWithinRaft[ccdName]['offset']
-                    yaw = geometryWithinRaft[ccdName]['yaw']
+                    yaw = geometryWithinRaft[ccdName]['yaw'] + raftYaw
                 else:
                     doffset = (0.0, 0.0,)
                     yaw = None
 
-                print(indent(), "%s_%s : " % (raftName, ccdName), file=fd)
+                print(indent(), "%s_%s : " % (raftName.split('_')[0], ccdName), file=fd)
```
timj: Is this change still needed? I see that in the commit history the name bounces from R00_W to R00W so I'm not sure if the splitting here is any longer relevant.

PR #152, opened 2019-12-13
DM-22550: Fix some header translations for LATISS data

timj:
@tribeiro, @patrickingraham Can you please take a look at this. It looks like a large review but in reality it's fairly small. Please focus on the changes in latiss.py and filters.py.

There are correction files that fix the disperser->diffuser filter problem.

The important things are:

* What should we have for the LATISS filters?
* Are my corrections in fix_header reasonable?
* Are you okay with the fixes to observation_type and physical_filter?

Finally, should I try to set DARKTIME to EXPTIME if it dark time is 0.0s but exp time is > 0?

Requires lsst/obs_base#190


python/lsst/obs/lsst/filters.py
```diff
@@ -53,3 +53,37 @@
                      abstract_filter="y",
                      lambdaEff=971.68, lambdaMin=975.0, lambdaMax=1075.0, alias=['y4']),
 )
+
+LATISS_FILTER_DEFINITIONS = FilterDefinitionCollection(
+    FilterDefinition(physical_filter="NONE",
+                     lambdaEff=0.0,
+                     alias={"no_filter", "OPEN"}),
+    FilterDefinition(physical_filter="275CutOn",
```
patrickingraham: where does this come from? I've never heard of it.

python/lsst/obs/lsst/filters.py
```diff
@@ -53,3 +53,37 @@
...
+                     lambdaEff=0.0),
```
patrickingraham: is 0.0 really the appropriate value?  There is no information about the filter wavelength other than what is in the title...

python/lsst/obs/lsst/filters.py
```diff
@@ -53,3 +53,37 @@
...
+    FilterDefinition(physical_filter="550CutOn",
+                     lambdaEff=0.0),
+    # The LSST Filters from L. Jones 04/07/10
```
patrickingraham: These are old but probably close enough for now. We'll have ~5 others to add (selected yesterday) that we'll use for early commissioning.

python/lsst/obs/lsst/translators/latiss.py
```diff
@@ -145,6 +152,70 @@ def can_translate(cls, header, filename=None):
...
+        modified = `bool`
+            Returns `True` if the header was updated.
+
+        Notes
+        -----
+        This method does not apply per-obsid corrections.  The following
+        corrections are applied:
+
+        * On June 24th 2019 the detector was changed from ITL-3800C-098
+          to ITL-3800C-068.  The header itself was not fixed until ???.
```
patrickingraham: I'd suggest putting a date or modifying the sentence to remove the ???

python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -443,7 +443,7 @@ def to_observation_type(self):
         obstype = self._header["IMGTYPE"]
         self._used_these_cards("IMGTYPE")
         obstype = obstype.lower()
-        if obstype == "skyexp":
+        if obstype in ("skyexp", "object"):
             obstype = "science"
```
patrickingraham: I'm assuming that these all get capitalized later?

timj: No, downstream they are always lower case in Gen3 and astro_metadata_translator. Gen2 forces them to upper case later...

python/lsst/obs/lsst/translators/latiss.py
timj: I don't know the date. I wanted that sentence to be there to remind us all that I am not meant to be forcing an override value here for the rest of time. Maybe replace with:

> The header is intended to be correct in the future.

?

python/lsst/obs/lsst/filters.py
timj: 0.0 seems to be our default for "no idea". This was in the LSSTCam filter list.

python/lsst/obs/lsst/filters.py
timj: I think these are used on BOT so I can remove them from LATISS list.

patrickingraham: the 550CutOn can be removed too.

mfisherlevine: Yup, these are BOT/TS8/TS3 filters.

python/lsst/obs/lsst/filters.py
mfisherlevine: They're also all set to zero because a single number tells you almost nothing about a filter. Left edge and right edge, or central and width, but a single number could be 1Å width filter, or 500nm. Given that that number isn't actually used anywhere either, as all the filter stuff was being overhauled, they were all just set to zero.

PR #158, opened 2020-01-16
DM-23090: Add LATISS commissioning filters

timj:
First attempt at adding the 4 commissioning filters. I've completely replaced the previous LSST filters that were in the LATISS definition.

Also some minor additional code cleanups since I was looking.


python/lsst/obs/lsst/filters.py
```diff
@@ -58,26 +58,17 @@
     FilterDefinition(physical_filter="NONE",
                      lambdaEff=0.0,
                      alias={"no_filter", "OPEN"}),
-    # The LSST Filters from L. Jones 04/07/10
-    FilterDefinition(physical_filter="u",
-                     abstract_filter="u",
-                     lambdaEff=364.59, lambdaMin=324.0, lambdaMax=395.0),
-    FilterDefinition(physical_filter="g",
+    FilterDefinition(physical_filter="blank_bk7_wg05",
+                     lambdaEff=0.0),
```
patrickingraham: if lambdaEff=0.0 are lambdaMin and lambdaMax not needed?

doc/lsst.obs.lsst/index.rst
```diff
@@ -24,6 +24,23 @@ You can find Jira issues for this module under the `obs_lsst <https://jira.lsstc
    adding-a-camera.rst
    testing.rst
 
+.. _lsst.obs.lsst-scripts:
```
patrickingraham: oh sure, jam something in here that's unrelated :)
It's documentation so I'll assume the content is good.

python/lsst/obs/lsst/lsstCamMapper.py
```diff
@@ -144,8 +144,8 @@ def __init__(self, inputPolicy=None, **kwargs):
         LsstCamMapper._nbit_patch = 5
         LsstCamMapper._nbit_filter = 6
 
-        LsstCamMapper._nbit_id = 64 - (LsstCamMapper._nbit_tract + 2*LsstCamMapper._nbit_patch +
-                                       LsstCamMapper._nbit_filter)
+        LsstCamMapper._nbit_id = 64 - (LsstCamMapper._nbit_tract + 2*LsstCamMapper._nbit_patch
```
patrickingraham: This is rather unclear to me what this is doing...
There is a limit on how many filters you can declare to afw?

python/lsst/obs/lsst/filters.py
timj: That seems to be the implication from looking elsewhere.

doc/lsst.obs.lsst/index.rst
timj: Yes. I was assuming you wouldn't worry to much.

python/lsst/obs/lsst/lsstCamMapper.py
timj: It's not obvious but the important thing for this pull request is that I moved the plus from the end of one line to the start of the next because of an RFC that got adopted yesterday.

doc/lsst.obs.lsst/index.rst
patrickingraham: tisk tisk.

python/lsst/obs/lsst/lsstCamMapper.py
patrickingraham: OK, please link the RFC to the ticket prior to merging.

doc/lsst.obs.lsst/index.rst
timj: It's turning on documentation for the command line scripts. It's not writing docs as such. I simply noticed that we hadn't put the right lines in the index file to turn on the documentation.

PR #159, opened 2020-01-17
DM-21308: Replace doApplyUberCal with doApplyExternal configs

erykoff:


config/compareWarpAssembleCoadd.py
```diff
@@ -30,4 +30,5 @@
 config.assembleStaticSkyModel.subregionSize = (10000, 200)
 
 # FUTURE Set to True if we run meas_mosaic or jointcal
```
mrawls: Please clarify this comment refers to 2 lines below, not just 1.

config/makeCoaddTempExp.py
```diff
@@ -31,7 +31,8 @@
 config.modelPsf.defaultFwhm = 7.7
 
 # FUTURE: Set to True when we decide to run jointcal
```
mrawls: Same as above, this comment applies to 2 configs and not just 1.

PR #160, opened 2020-01-22
DM-23149: Add support for on sky LATISS data

timj:


python/lsst/obs/lsst/translators/latiss.py
```diff
@@ -223,11 +225,27 @@ def fix_header(cls, header):
...
+                modified = True
+
         if header.get("SHUTTIME"):
             log.debug("%s: Forcing SHUTTIME header to be None", obsid)
             header["SHUTTIME"] = None
             modified = True
 
+        if "OBJECT" not in header:
+            # Only patch OBJECT IMGTYPE
+            if "IMGTYPE" in header and header["IMGTYPE"] == "OBJECT":
```
mfisherlevine: Presumably this plays correctly with the `OBJECT_IS_ENGTEST` override stuff above? If not, does it matter?

python/lsst/obs/lsst/translators/latiss.py
```diff
@@ -260,13 +278,6 @@ def compute_detector_exposure_id(exposure_id, detector_num):
             log.warning("Unexpected non-zero detector number for LATISS")
         return exposure_id
 
-    @cache_translation
-    def to_location(self):
```
mfisherlevine: I assume there was a reason for removing this, right?

python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -17,13 +17,16 @@
...
 
+import astropy.coordinates
 import astropy.units as u
 from astropy.time import Time, TimeDelta
 from astropy.coordinates import EarthLocation
 
 from lsst.utils import getPackageDir
 
 from astro_metadata_translator import cache_translation, FitsTranslator
+from astro_metadata_translator.translators.helpers import tracking_from_degree_headers, \
```
mfisherlevine: I know some people care deeply about using parens for line continuation on imports, but I'm not one of them, and also know that `astro_metadata_translator` doesn't play by all the same rules.

python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -514,3 +533,31 @@ def to_physical_filter(self):
...
+            radec = tracking_from_degree_headers(self, radecsys, radecpairs)
+
+        return radec
+
+    @cache_translation
+    def to_altaz_begin(self):
+        if not self.is_science_on_sky():
+            return None
+
+        return altaz_from_degree_headers(self, (("ELEND", "AZEND"),),
```
mfisherlevine: The function is `to_altaz_begin` but the method being called seems to use the end values, not the begin values. Is that correct?

python/lsst/obs/lsst/translators/latiss.py
timj: Because I'm now using the default implementation from the base class.

python/lsst/obs/lsst/translators/latiss.py
timj: Yes. By this point IMGTYPE won't be OBJECT in that case.

python/lsst/obs/lsst/translators/lsst.py
timj: Hmm, am I breaking policy? I can easily change it.

python/lsst/obs/lsst/translators/lsst.py
timj: Well spotted. This is from me debugging and trying to work out why the RA/DEC was not matching the AZEL. Will fix.

python/lsst/obs/lsst/translators/lsst.py
mfisherlevine: Policy? I don't think so, no. I just know that some people feel strongly, but like I said, I'm not one of them, so do whatever you like really 🙂

python/lsst/obs/lsst/translators/lsst.py
mfisherlevine: Glad it wasn't all in vain 👍

PR #161, opened 2020-01-24
DM-23171: Add exposure_group calculation and make visit_id use it

timj:
Exposure group now read from GROUPID header if defined,
else uses the exposure_id as a string. Visit has to be
changed to match since things with the same exposure_group
must now have matching visit id.

If exposure_group is a number use it, else derive it
from the group ID.

Depends on lsst/astro_metadata_translator#35


python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -515,8 +522,62 @@ def to_exposure_id(self):
...
+            if n is not None:
+                n = int(n)
+            else:
+                n = 0
+            iso = datetime.datetime.strptime(iso_str, "%Y-%m-%dT%H:%M:%S")
+
+            tdelta = iso - TZERO_DATETIME
+            epoch = int(tdelta.total_seconds())
+
+            # Form the integer from EPOCH + 3 DIGIT FRAC + 0-pad N
```
SimonKrughoff: You don't mean addition here.  It's an append.  I agree it's clear from the next line.

python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -30,7 +33,11 @@
 
 # LSST day clock starts at UTC+8
 ROLLOVERTIME = TimeDelta(8*60*60, scale="tai", format="sec")
-TZERO = Time("2010-01-01T00:00", format="isot", scale="utc")
+TZERO = Time("2015-01-01T00:00", format="isot", scale="utc")
+TZERO_DATETIME = TZERO.to_datetime()
+
+# Regex to use for parsing a GROUPID string
```
SimonKrughoff: Maybe add an example string this is meant to match?

python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -515,8 +522,62 @@ def to_exposure_id(self):
...
+            # using a hash function. Use the first N hex digits
+            group_bytes = exposure_group.encode("us-ascii")
+            hasher = hashlib.blake2b(group_bytes)
+            # Need to be big enough it does not possibly clash with the
+            # date-based version above
+            digest = hasher.hexdigest()[:14]
+            visit_id = int(digest, base=16)
+
+            # To help with hash collision, append the string length
+            visit_id = int(f"{visit_id}{len(exposure_group):02d}")
```
SimonKrughoff: Why truncate to two digits?

python/lsst/obs/lsst/translators/lsst.py
timj: Yes. Good point.

python/lsst/obs/lsst/translators/lsst.py
timj: The length comes from a FITS header value and unless we are going crazy with CONTINUE header cards 72 is the longest group ID.

PR #164, opened 2020-01-31
DM-23222: More LATISS header fixes

timj:


python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -617,7 +624,11 @@ def to_tracking_radec(self):
 
     @cache_translation
     def to_altaz_begin(self):
-        if not self.is_science_on_sky():
+        if not self._is_on_mountain():
+            return None
+
+        # ALTAZ always relevant unless bias or dark
+        if self.to_observation_type() in ("bias", "dark"):
```
patrickingraham: or flat

timj: We take flats with the shutter open at fixed AZEL so we want the translator to report that AZEL. Isn't that right?

patrickingraham: It certainly can... but it could with bias and dark as well... 
I think it doesn't matter really. It'll never get used.

patrickingraham: i'm fine with reporting it.

PR #171, opened 2020-02-07
Remove refCols from mapper policies.

ktlim:
Also clean up some spurious calibDates.

### DIFF
--- a/policy/lsstCamMapper.yaml
+++ b/policy/lsstCamMapper.yaml
@@ -83,8 +83,6 @@ calibrations:
     level: Ccd
     persistable: BaseCatalog
     python: lsst.meas.algorithms.simple_curve.Curve
-    refCols:
-    - visit
     storage: FitsCatalogStorage
     tables: qe_curve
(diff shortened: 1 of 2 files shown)


policy/lsstCamMapper.yaml
```diff
@@ -79,12 +79,11 @@ calibrations:
   qe_curve:
     columns:
     - detector
-    - calibDate
+    - dateObs
```
SimonKrughoff: I was under the impression these were the columns in the calib repository to find the calib in question.  I must be wrong about that, but wanted to bring it up.

ktlim: These are columns in the _main_ registry that are used to get values to look up in the calib registry.  The `obsTimeName` entry below is the name of the field whose value is compared against the `(validStart, validEnd)` interval.

ktlim: This is unfortunately not well-documented, but there is a code comment to this effect at https://github.com/lsst/obs_base/blob/master/python/lsst/obs/base/mapping.py#L490-L491

SimonKrughoff: OK.  Thanks for the clarification.

PR #172, opened 2020-02-07
Revert "Merge pull request #171 from lsst/tickets/DM-23397"

ktlim:
This reverts commit 7b8605421759c4458cdcb9bdcd5caea07397a034, reversing
changes made to 3a1ca26540ee34009b5b5f5abb0e5cb91984dcac.


ktlim:
This is necessary because the proposed change breaks the imsim test and also other uses of extended data ids; it will have to be rethought.

PR #176, opened 2020-02-13
Template and config updates for defects

mfisherlevine:
Updates so the defect finding task actually works, and is compatible with ingest.

### DIFF
--- a/policy/lsstCamMapper.yaml
+++ b/policy/lsstCamMapper.yaml
@@ -610,4 +610,4 @@ datasets:
     storage: FitsStorage
     tables:
     - raw
-    template: calibrations/defects/defects-det%(detector)03d.fits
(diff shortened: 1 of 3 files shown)


policy/lsstCamMapper.yaml
```diff
@@ -610,4 +610,4 @@ datasets:
     storage: FitsStorage
     tables:
     - raw
-    template: calibrations/defects/defects-det%(detector)03d.fits
+    template: calibrations/SomethingHasToGoHereButThisIsOverwrittenLater%(detector)03d.extension
```
czwa: I would prefer a different string, but agree that this makes it easy to search in case something goes wrong.

mfisherlevine: Yeah, it's gross. Talking to KT about this and other things soon. Won't merge till that's happened at least. (Mostly made it like that to force us to discuss this anyway!)

PR #180, opened 2020-02-20
DM-23589: Add 3 new filters to LATISS

timj:


python/lsst/obs/lsst/lsstCamMapper.py
```diff
@@ -142,7 +142,7 @@ def __init__(self, inputPolicy=None, **kwargs):
 
         LsstCamMapper._nbit_tract = 16
         LsstCamMapper._nbit_patch = 5
-        LsstCamMapper._nbit_filter = 6
+        LsstCamMapper._nbit_filter = 7
```
mfisherlevine: This give us 128. Is this one per filter as defined above, or one per combination? If the former, 128 should be enough. If the latter, I suspect this should change to at least 8.

Also, does changing this mean that things need reingesting as numbers change, or no?

timj: Looks like this is mainly used for source IDs from coadds.

timj: Oh and remember that every time we add a filter we add that filter and that filter with each grating combination. I added three filters here so that's an extra 18 filters. (there are 5 gratings defined including the empties and none and ???).

PR #181, opened 2020-02-21
DM-22655: Genericize gen2to3.py to be useable with any gen2 repo

parejkoj:


parejkoj:
I've created DM-23778 for configuring and testing obs_lsst conversion. I have questions there about what testdata we have to work with.

PR #184, opened 2020-03-09
Edited comCam yaml file with realistic gains

craiglagegit:
Edited the policy/comCam/R22.yaml file with gains and noise values as extracted on the comCam CCDs with the ptc.py code in cp_pipe.  Since we will probably want to do this again, I wrote a script in /project/cslage/ComCam/notebooks/Write_PTC_Gains_to_YAML_09Mar20.ipynb that can read the stored gains and update the yaml file.  I tested the updated gains on the pinhole images using the notebook at /project/cslage/ComCam/notebooks/Test_Pinhole_Image_09Mar20.ipynb.  Only the one yaml file has been changed.


craiglagegit:
I'm closing this pull request because there are two other things that need to be fixed, so we might as well do them all at once:
(1) Also assign relaistic gains to the AuxTel CCD
(2) The ComCam CCDs are not assembled correctly within the raft.  This needs to be fixed as well.

PR #189, opened 2020-03-20
DM-23778: Add gen2 to 3 repo conversion

timj:


policy/comCam/comCamMapper.yaml
```diff
@@ -58,7 +58,7 @@ exposures:
   deepCoadd_psfMatchedWarp:
     template: deepCoadd/%(filter)s/%(tract)d/%(patch)s/psfMatchedWarp-%(filter)s-%(tract)d-%(patch)s-%(expId)d.fits
   dcrCoadd_directWarp:
-    template: dcrCoadd/%(filter)s%(subfilter)s/%(tract)d/%(patch)s/warp-%(filter)s%(subfilter)s-%(tract)d-%(patch)s-%(expId)d.fits
+    template: dcrCoadd/%(filter)s%(subfilter)d/%(tract)d/%(patch)s/warp-%(filter)s%(subfilter)d-%(tract)d-%(patch)s-%(expId)d.fits
```
TallJimbo: I didn't realize subfilters were always numbers.  @isullivan , can you confirm?

config/convertRepo.py
```diff
@@ -0,0 +1,12 @@
+# Config overrides for converting gen2 to gen3 repos.
+
+from lsst.obs.base.gen2to3 import Translator, AbstractToPhysicalFilterKeyHandler
```
TallJimbo: This import is for side-effects?

tests/test_convert2to3.py
```diff
@@ -0,0 +1,113 @@
+# This file is part of obs_subaru.
```
TallJimbo: Copy/paste relic.

tests/test_convert2to3.py
```diff
@@ -0,0 +1,113 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Unit tests for HSC gen2 to gen3 conversion.
+"""
```
TallJimbo: Copy/paste relic.  Not sure a module docstring is worthwhile here anyway.

python/lsst/obs/lsst/gen3/instrument.py
```diff
@@ -166,25 +166,40 @@ def writeCuratedCalibrations(self, butler):
...
+                         "storageClass": "QECurve"},
+        }
+
+        for typeName, definition in curatedCalibrations.items():
+            # We need to define the dataset types.
+            datasetType = DatasetType(typeName, definition["dimensions"],
+                                      definition["storageClass"],
+                                      universe=butler.registry.dimensions)
+            butler.registry.registerDatasetType(datasetType)
+            self._write_obs_lsst_data(butler, datasetType)
```
TallJimbo: I'm not really convinced that making this code generic over dataset types is the right call; smells a little like incidental duplication.  But I guess we'll see when the third dataset type is added here: if that requires a change to the generic version, we should just split it into separate blocks for each dataset type to let them do their own thing.

policy/comCam/comCamMapper.yaml
timj: Hmm. Okay -- I was getting inconsistency errors from the conversion tool because other templates in obs_base (I think) define it as a number and not a string. Now that I look in daf_butler subfilter has to be a string so does that mean I need to fix obs_base?

python/lsst/obs/lsst/gen3/instrument.py
timj: Remember that obs_*_data packages are by definition being set up in the same way such that the read_all function will return them in standard form for a given dataset type. If someone is using obs_*_data calibrations then this code will be the same in every obs package -- it's already the same between obs_subaru and obs_lsst and will be the same for obs_decam.  Ideally obs_base would have a mode that scans the obs_*_data package for curated calibrations from a standardized set and the obs packages don't have to worry about _data at all.  The reason ingestCuratedCalibrations works for gen2 is because of this standardization.  The interesting bit is going to come when we have a curated calibration that has different gen3 dimensions.

config/convertRepo.py
timj: No, it's a copy and paste from obs_decam which added some explicit translators in the config itself.

policy/comCam/comCamMapper.yaml
timj: To be explicit here. obs_base uses:
```
%(filter)s%(subfilter)dof%(numSubfilters)d/
```
obs_decam uses:
```
%(filter)s%(subfilter)d
```
obs_lsst uses `%(filter)s%(subfilter)s` and daf_butler defines subfilter to be a string. Since the conversion tool reads the templates from obs_base you get this error:
```
ValueError: Multiple types for key 'subfilter': <class 'str'> (from dcrCoadd/%(filter)s%(subfilter)s/%(tract)d/%(patch)s/warp-%(filter)s%(subfilter)s-%(tract)d-%(patch)s-%(expId)d.fits) vs. <class 'int'>.
```
So I'm not entirely sure how to proceed. What I did do was make all the obs packages consistent but that is not consistent with gen3. Do I merge what I've got and we resolve this on a new ticket?

TallJimbo: I don't know what the right fix is, so I guess I'd vote for changing as little as possible to get things working until we know more, but your guess is as good as mine on how best to accomplish that.

PR #190, opened 2020-03-24
DM-23975: Make subfilter consistent between Gen 2 and 3

isullivan:
The subfilter dimension needs to be a string in Gen 3, so for consistency it should also be a string in Gen 2 even though an int would be more natural.


timj:
Closing since we've decided that the real fix is in lsst/daf_butler#242

PR #197, opened 2020-04-10
DM-15536: adapt to moving visit definition out of ingest

TallJimbo:
The tests here now just test that trying to define visits from the test data already included in the repo correctly does nothing, because (apparently) none of the exposures here have `observation_type = 'science'`.  When that changes someday in the future (and we include those data IDs in the ingest tests), the tests will fail unless we also set the "visits" attribute to a non-empty dictionary of expected visits, as in the obs_subaru and obs_decam tests.

### DIFF
--- a/config/convertRepo.py
+++ b/config/convertRepo.py
@@ -10,3 +10,6 @@ config.ccdKey = "detector"
 config.datasetIgnorePatterns.extend(["_raw", "raw_hdu", "raw_amp"])
 
 config.curatedCalibrations.extend(["defects", "qe_curve"])
+
+import os
+config.defineVisits.load(os.path.join(os.path.dirname(__file__), "defineVisits.py"))
--- a/config/defineVisits.py
+++ b/config/defineVisits.py
@@ -1,0 +1,1 @@
+config.groupExposures.name = "by-group-metadata"


config/convertRepo.py
```diff
@@ -10,3 +10,7 @@
 config.datasetIgnorePatterns.extend(["_raw", "raw_hdu", "raw_amp"])
 
 config.curatedCalibrations.extend(["defects", "qe_curve"])
+
+import os
+from lsst.utils import getPackageDir
+config.defineVisits.load(os.path.join(getPackageDir("obs_lsst"), "config", "defineVisits.py"))
```
timj, line 16: You do not need to use getPackageDir here. Instead you can use the new `__file__` since you are wanting to read the `defineVisits.py` file from the same directory that this config file is located.

TallJimbo: `getPackageDir` seems to be a pretty strong convention in obs package config files - especially in obs_subaru, which has many more, but here as well - and I'm wary of going against it without explicitly RFC'ing or at least broadly announcing an intent to move away from it.

timj: It's a strong convention because until February (DM-23359) you couldn't use `__file__` -- we had a long discussion on Slack back then where we realized that the same logic that led us to specify test data relative to the tests works for config files where we absolutely want to find the config file in the same directory of another config file. getPackageDir is a hack in that context but it's the only thing we could do previously.

timj: Oh and there was a [community post](https://community.lsst.org/t/pex-config-configs-now-know-where-they-are/4023).

TallJimbo: I started an across-the-board conversion of all three Gen3 obs packages to use file `__file__`, but I've decided that's enough that it merits its own ticket.  I've put all that on [DM-24495](https://jira.lsstcorp.org/browse/DM-24495), which is now out for review, and if that merges first I'll modify this branch to use the new style.  If it merges second I'll fix the DM-24495 branch.

PR #201, opened 2020-04-16
DM-24495: Switch to using __file__ to reference other configs.

TallJimbo:
See https://community.lsst.org/t/pex-config-configs-now-know-where-they-are/4023.

### DIFF
--- a/config/sky.py
+++ b/config/sky.py
@@ -22,11 +22,6 @@
 
 import os.path
 
-# The following will have to be uncommented and adapted when the
-# X-talk correction will be active
-#from lsst.obs.lsst.isr import SubaruIsrTask
-#config.isr.retarget(SubaruIsrTask)
-#config.isr.load(os.path.join(getPackageDir("obs_lsst"), "config", "isr.py"))
 configDir = os.path.dirname(__file__)
 
 config.load(os.path.join(configDir, "lsstCamCommon.py"))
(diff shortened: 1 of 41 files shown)


config/sky.py
```diff
@@ -22,13 +22,12 @@
 
 import os.path
 
-from lsst.utils import getPackageDir
 # The following will have to be uncommented and adapted when the
 # X-talk correction will be active
 #from lsst.obs.lsst.isr import SubaruIsrTask
 #config.isr.retarget(SubaruIsrTask)
 #config.isr.load(os.path.join(getPackageDir("obs_lsst"), "config", "isr.py"))
```
czwa, line 29: Should this be updated as well?  I don't know the logic behind this entire commented block (which seems thoroughly out of date), but if getPackageDir will no longer be imported, this line will fail if uncommented.

TallJimbo: I wasn't sure what to do with it either.  `obs_subaru/config/hsc/sky.py` does indeed load its own ISR config override, but I don't know enough about what this task does to know why it might be okay that we don't do that right now for obs_lsst - I suspect the answer is that we never run this task on any LSST camera, and hence this whole file is a waste of time.  But I don't know.

I suppose the path of least resistance is just to switch to using `__file__` in the comment, too.

TallJimbo: I looked at this again and realized that this config _already_ loads obs_lsst's custom ISR, so I'm pretty confident the entire comment is defunct.  I'm just removing it.

PR #203, opened 2020-04-27
DM-24517: Reading of raws from existing imsim & phosim repos broken by DM-23584

laurenam:


.gitignore
```diff
@@ -17,4 +17,6 @@ ts8/CALIB/calibRegistry.sqlite3
 ts8/CALIB/qe_curve/
 lsstcam/CALIB/calibRegistry.sqlite3
 lsstcam/CALIB/qe_curve/
-
+data/input/phosim/calibRegistry.sqlite3
```
timj, line 20: I really don't understand why these files appear so I worry that ignoring them makes it easy to forget there is a possible problem somewhere.

laurenam: Yeah, it didn't seem like they should be there to me either, but I didn't dig deep enough to figure out where they're coming from.  I could make a ticket for someone who knows better to look into it. Shall I leave this in to avoid the annoyance for now but update the commit message (or comment in the `.gitignore` file directly?) to indicate they should be removed with that ticket...or should I just get rid of it altogether?

timj: I finally did what I should have done ages ago and make the data directory read only. The new files are created because of a bug in the gen2to3 code. Please don't merge the .gitignore change. I'll make a new ticket for the problem.

laurenam: Sounds good.  I'll remove that commit and restart Jenkins to be sure...and it looks like the current one had an unrelated failure on centos-6 (osx passed and centos-7 made it past the failure point)? https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/31688/pipeline/48

timj: That ip_isr error is weird. I wouldn't insist on doing a whole new jenkins run for a .gitignore change since .gitinore is not anything that's tested during the build

laurenam: Well...I'm overly paranoid, so it's already going (and can't hurt 😉).  Thanks again!

PR #206, opened 2020-05-01
DM-24664: Refactor data ID translation rule system

TallJimbo:


python/lsst/obs/lsst/instrument.py
```diff
@@ -151,6 +152,12 @@ def extractDetectorRecord(self, camGeomDetector):
             raft=group,
         )
 
+    def makeDataIdTranslatorFactory(self) -> TranslatorFactory:
+        # Docstring inherited from lsst.obs.base.Instrument.
+        factory = TranslatorFactory()
+        factory.addGenericInstrumentRules(self.getName(), detectorKey="detector", exposureKey="expId")
```
parejkoj, line 158: I was thrown off by `detectorKey="detector"` (aren't they all?!?) until I remembered that this was about the names in gen2. Probably not worth a comment though.

PR #211, opened 2020-05-14
DM-24908: Restore visit in place of expId for all templates except raw

erykoff:


policy/imsim/imsimMapper.yaml
```diff
@@ -4,88 +4,97 @@ exposures:
   raw_amp:
     template: raw/%(run)s/%(raftName)s/%(expId)08d-%(raftName)s-%(detectorName)s-det%(detector)03d.fits[%(channel)d]
   postISRCCD:
-    template: postISRCCD/%(expId)08d-%(filter)s/%(raftName)s/postISRCCD_%(expId)08d-%(filter)s-%(raftName)s-%(detectorName)s-det%(detector)03d.fits
+    template: postISRCCD/%(visit)08d-%(filter)s/%(raftName)s/postISRCCD_%(visit)08d-%(filter)s-%(raftName)s-%(detectorName)s-det%(detector)03d.fits
```
TallJimbo, line 7: Could we revert this one (and perhaps the next, for consistency)?

Maybe I should asking @yalsayyad to debate this more directly, as I'm not sure @erykoff cares, but Gen3 `IsrTask` and `CharacterizeImageTask` are already consistent in calling this `exposure` (the Gen3 version of `expId`).  I don't think we can easily change the Gen3 tasks to use `visit` for `postISRCCD` without breaking Gen3 `IsrTask` usage in CPP.  That doesn't need to be affected by this change, but since we can just which concept we use here, it seems like we might as well be consistent.

policy/imsim/imsimMapper.yaml
```diff
@@ -4,88 +4,97 @@ exposures:
...
+    template: focal_plane/%(dstype)s-v%(visit)08d.fits
   focal_plane_png:
-    template: focal_plane/%(dstype)s-v%(expId)08d.png
+    template: focal_plane/%(dstype)s-v%(visit)08d.png
   binned_sensor_fits:
-    template: binned_sensor/%(dstype)s-v%(expId)08d/%(raftName)s/%(detectorName)s.fits
+    template: binned_sensor/%(dstype)s-v%(visit)08d/%(raftName)s/%(detectorName)s.fits
   binned_sensor_fits_halves:
-    template: binned_sensor/%(dstype)s-v%(expId)08d/%(raftName)s/%(detectorName)s_%(half)s.fits
+    template: binned_sensor/%(dstype)s-v%(visit)08d/%(raftName)s/%(detectorName)s_%(half)s.fits
```
TallJimbo, line 100: I'm not sure about these `focal_plane_*` and `binned_sensor_*` ones - I don't know what they're for, but we should try to find out, because I think it's possible they're more about raw data than processed data.

policy/imsim/imsimMapper.yaml
erykoff: I could change it, sure, if that will make you happy.  But this is for Gen2 ImSim, where it really doesn't matter.  But if we don't want to start a precedent, I'll leave it at expId.

policy/imsim/imsimMapper.yaml
erykoff: Interesting point.  I'll figure this out before merging.  (But again, these are equivalent for Gen2 ImSim.)

erykoff: I have confirmed that the code comes from https://github.com/LSSTDESC/desc_sim_utils/blob/master/bin.src/makeFpSummary.py#L120 and this is calexp-level binning, so `visit` is appropriate.

PR #224, opened 2020-06-16
DM-25171: Change how headers are read from raw data in gen2

timj:
* Add helper function that reads the primary header and merges it with the first amp header. PhoSim requires this.
* Add specialist bypass_raw_hdu so that assembled raws can use the helper function.
* Add specialist bypass_raw_md method that uses the same helper function.  This allows header corrections  to be applied.
* Add new tests to ensure that visitInfo, Exposure, and metadata have header corrections applied.

### DIFF
--- a/python/lsst/obs/lsst/_fitsHeader.py
+++ b/python/lsst/obs/lsst/_fitsHeader.py
@@ -1,0 +1,71 @@
+# This file is part of obs_lsst.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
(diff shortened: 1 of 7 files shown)


python/lsst/obs/lsst/_fitsHeader.py
```diff
@@ -0,0 +1,72 @@
...
+            hdu = requested
+
+    if hdu is not None:
+        md = lsst.afw.fits.readMetadata(fileName, hdu=hdu)
+    else:
+        # For raw some of these files need the second header to be
+        # read as well
+        phdu = lsst.afw.fits.readMetadata(fileName, 0)
+        md = lsst.afw.fits.readMetadata(fileName)
+        if "INHERIT" not in phdu:
```
SimonKrughoff, line 67: The implication is that if `INHERIT` is in the primary header unit, the FITS reading code should already bring in keys from other extensions?

tests/data/LSST-PhoSim-204595.yaml
```diff
@@ -0,0 +1,2 @@
+NEWHDR: corrected
+TEMPERA: 10.0
```
SimonKrughoff, line 2: Don't you need the other test values like `PRESS` and `AIRMASS`?

SimonKrughoff: Oh, I see the pressure being tested directly from the metadata, where the temperature is being tested from `getWeather`.  I'll assume I'm just missing some detail about why they are tested in different ways.

python/lsst/obs/lsst/_fitsHeader.py
timj: Well spotted. INHERIT never goes in primary header so this test is irrelevant.

tests/data/LSST-PhoSim-204595.yaml
timj: Yes. This is for the corrections application. I'm checking that the new header turns up (so corrections are being applied for raw_md) and also that the VisitInfo gets corrected headers (so that raw_visitInfo is updated). I think that's sufficient to show that corrections are being applied in all three cases.

PR #235, opened 2020-07-27
outdated: fix amp-order for lsstCam readnoise, gain, and fullwell

bxin:
In importing the read noise, gain, and saturation as measured by the camera team, incorrect amp-ordering was assumed for data out of the eTraveler. The correct order should be C10,...,17,07,...,00. This is now fixed.

The PR only changes the numbers in policy/lsstCam/Rxx.yaml
by re-assigning the read noise, gain, and saturation values to the correct amps.

`bin.src/update_raft.py` is a script we used to help with this update. See https://jira.lsstcorp.org/browse/DM-22824


timj:
As a general comment we are trying not to have large amount of code in bin scripts. The "business logic" should go in the python/lsst/obs/lsst/script directory and the bin.src file should be a couple of lines long.

timj:
Oh, and if the script is there for historical reasons because that's what you used to fix things up, but is not useful longer term it should probably be in a different place so it doesn't confuse the path.

timj:
I'm sorry for writing lots of little comments: Is there a Jira ticket for this so we can use a proper branch name?

If update_raft is staying in bin.src it would be better to have a less confusing name since all users of obs_lsst will see it in their path and `--help` doesn't give much information as to what it does.

timj:
Can we also please reference a Jira ticket for this work?

bxin:
help needed: I’ve renamed the branch from u/bxin/fixQEAmpOrder to tickets/SE-1493. What do I do with the PR? Do I need a new PR? 
GH doesn’t seem to let me change the `from` field of the PR.

timj:
You need to close this PR and create a new one.

PR #238, opened 2020-08-01
Tickets/dm 26205 Fix obs_lsst to handle corner raft data

RobertLuptonTheGood:


python/lsst/obs/lsst/script/generateCamera.py
```diff
@@ -195,23 +195,34 @@ def indent():
 
             try:
                 detectorType = raftCcdData["detectorType"]
+            except KeyError:
+                raise RuntimeError("Unknown detector type %s" % detectorType)
```
parejkoj, line 199: I'm pretty sure these lines do not do what you intended them to do. Did you mean `raftCcdData[detectorType]`? And is there a test for this?

PR #240, opened 2020-08-03
DM-26205: Fix obs_lsst to handle corner raft data

RobertLuptonTheGood:
Second attempt

### DIFF
--- a/python/lsst/obs/lsst/lsstCamMapper.py
+++ b/python/lsst/obs/lsst/lsstCamMapper.py
@@ -302,7 +302,27 @@ class LsstCamBaseMapper(CameraMapper):
         ValueError
             The channel number requested in ``dataId`` is out of range.
         """
-        nChannel = 16                   # number of possible channels, 1..nChannel
+        # set number of possible channels, 1..nChannel
+        # The wave front chips are special, 4k x 2k with only 8 amps
+
+        if "detectorName" in dataId:
+            detectorName = dataId.get("detectorName")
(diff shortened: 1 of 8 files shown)


python/lsst/obs/lsst/lsstCamMapper.py
```diff
@@ -302,7 +302,13 @@ def query_raw_amp(self, format, dataId):
         ValueError
             The channel number requested in ``dataId`` is out of range.
         """
-        nChannel = 16                   # number of possible channels, 1..nChannel
+        # set number of possible channels, 1..nChannel
+        # The wave front chips are special, 4k x 2k with only 8 amps
+        if dataId.get("detector") in [191, 192, 195, 196, 199, 200, 203, 204] or \
```
teweitsai, line 307: Where is the `detector` Id defined (any `yaml` file or a function to decide the `detector` Id)? Is it possible to get the corner detector ID programmatically (or by other functions) instead of the hard-coded IDs?

SimonKrughoff: I think I'd prefer to see something more like:
```if self.camera[dataId["detector"]].getName[-3:] in ("SW0", "SW1"):```

Since it's possible change the detector ID with changes to the yaml, those numbers are not guaranteed for all time.

RobertLuptonTheGood: Good idea.  I implemented a (more robust) form of this proposal.

PR #241, opened 2020-08-06
Fix astrometry for LATISS

mfisherlevine:


python/lsst/obs/lsst/assembly.py
```diff
@@ -52,10 +53,19 @@ def attachRawWcsFromBoresight(exposure):
     visitInfo = MakeRawVisitInfoViaObsInfo(logger)(md)
     exposure.getInfo().setVisitInfo(visitInfo)
 
+    # LATISS (and likely others) need flipping, DC2 etc do not
+    obsInfo = ObservationInfo(md)
```
timj, line 57: There is an implicit double calculation of the ObservationInfo that we don't need. I think if you move this line above the visitInfo line you can then do

```
visitInfo = MakeRawVisitInfoViaObsInfo.observationInfo2visitInfo(obsInfo, log=logger)
```

python/lsst/obs/lsst/assembly.py
```diff
@@ -63,6 +63,9 @@ def attachRawWcsFromBoresight(exposure):
         exposure.setWcs(createInitialSkyWcs(visitInfo, exposure.getDetector(), flipX=flipX))
         return True
 
+    if obsInfo.observation_type != "science":
```
ktlim, line 66: Are you sure you don't want `== "science"` here?

mfisherlevine: Haha, yes indeed. Sorry, this wasn't actually out for review, I only made the PR to make it easier for Tim to take a glance at the changes.

python/lsst/obs/lsst/assembly.py
mfisherlevine: Yup, for sure, I spotted that but hadn't got around to tidying up late last night and didn't know the right incantation off the top of my head.

python/lsst/obs/lsst/assembly.py
mfisherlevine: In moving this we've lost the ability to put the dataId in the message unless I change to pass it through. It won't need changing anywhere else, so doesn't really change anything, but it would only be for the message - is it worth it? I could default it to `None` and add it to the message if present as a compromise, but I'm curious about other people's thoughts on elegance/cruft in the code vs. completeness of error messages.

Personally I think that given it will be surrounded by other messages about that dataId it's not really necessary but certainly welcome opinions.

timj: Consider adding a `errmsg=None` optional parameter so that the caller can pass in the message with the dataId that can be appended to the generic message.

python/lsst/obs/lsst/assembly.py
```diff
@@ -27,11 +27,12 @@
...
 from lsst.geom import Box2I, Extent2I
 from lsst.ip.isr import AssembleCcdTask
+from astro_metadata_translator import ObservationInfo
 
-logger = lsst.log.Log.getLogger("LsstCamAssembler")
+logger = lsst.log.Log.getLogger("obs.lsst.assembly")
 
 
-def attachRawWcsFromBoresight(exposure):
+def attachRawWcsFromBoresight(exposure, dataIdForErrMsg=None):
```
timj, line 35: I prefer that this was simply errMsg and you formed the string in the caller. Then this routine is generic and doesn't require that a dataId be passed in.

python/lsst/obs/lsst/assembly.py
```diff
@@ -52,10 +53,19 @@ def attachRawWcsFromBoresight(exposure):
...
+        flipX = True
+
     if visitInfo.getBoresightRaDec().isFinite():
-        exposure.setWcs(createInitialSkyWcs(visitInfo, exposure.getDetector()))
+        exposure.setWcs(createInitialSkyWcs(visitInfo, exposure.getDetector(), flipX=flipX))
         return True
 
+    if obsInfo.observation_type == "science":
+        logger.warn("Unable to set WCS from header as RA/Dec/Angle are unavailable%s" %
+                    ("" if dataIdForErrMsg is None else " for dataId %s" % dataIdForErrMsg))
```
timj, line 68: Don't use `%` formatting in log calls. That fist `%` should be `,`

python/lsst/obs/lsst/assembly.py
mfisherlevine: Feels weirder to me to be formatting error messages that might not end up existing, no? That doesn't get logged at all if the wcs creation succeeds (most of the time) so making it up ahead of time feels more wrong to me than the alternative.

python/lsst/obs/lsst/assembly.py
mfisherlevine: `,`, `+` or simply continuation, yes. But I thought the second `%s` was _preferred_ for log formatting, no?

python/lsst/obs/lsst/assembly.py
timj: There is precedent for this. See for example fixAmpGeometry in that same file. It seems weirder to me that we pass in a very specific piece of information that is only useful for failure.

python/lsst/obs/lsst/assembly.py
timj: I meant the `%` after the first quote at the end of the line. It should be a `,` since you are meant to defer the formatting.

mfisherlevine: Agreed.

python/lsst/obs/lsst/assembly.py
mfisherlevine: I do not agree that that is a better solution. I think that both scattering callbacks or preemptively formatting up error message which might never appear are much worse than having an option parameter to improve an error message.

timj: Every single test in unittest takes a `msg=` parameter that only appears if the test fails. There is significant precedence for passing in a pre-formatted string on error. (I'm not a fan of that call back though). If you don't want it to be actually formatter it could be given as a tuple:
```
attachRawWhatever(exposure, msg=("Some dataiD issue: %s", dataId))
```
which could then be passed directly to the logger overriding the default message completely with `log.warn(**msg)`

mfisherlevine: Unit tests are in a different boat to production code, and all this seems like needless complexity tbh. I don't understand why formatting a string up front to pass in as an optional parameter to maybe be displayed on error is better than passing an optional dataId for use if needed.

timj: Ok, fine. Since this is gen2 specific I'm not going to fight.

PR #242, opened 2020-08-12
DM-23302: Add ci_cpp package to allow calibration product generation to be tested

czwa:
Modify LATISS calibration production pipelines to fix errors.

### DIFF
--- a/pipelines/latiss/cpDark.yaml
+++ b/pipelines/latiss/cpDark.yaml
@@ -13,7 +13,7 @@ tasks:
       doVariance: True
       doLinearize: False
       doCrosstalk: False
-      doDefect: False
+      doDefect: True
       doNanMasking: True
       doInterpolate: True
       doBrighterFatter: False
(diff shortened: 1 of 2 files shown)


pipelines/latiss/cpDark.yaml
```diff
@@ -13,7 +13,7 @@ tasks:
       doVariance: True
       doLinearize: False
       doCrosstalk: False
-      doDefect: False
+      doDefect: True
```
plazas, line 16: Why was this False before and now it is OK to have it as True? Same for the cpFlat.yaml file. I assume that this could be overwritten by the user, but wondered if there's any reason I'm missing.

czwa: The defects in obs_lsst weren't being ingested in gen3 when I wrote DM-21212.  They are now ingested, and so we can do defect masking before constructing the calibrations.

mfisherlevine (approved):

PR #243, opened 2020-08-17
SE-1493: fix amp-order for lsstCam readnoise, gain, and fullwell

bxin:
In importing the read noise, gain, and saturation as measured by the camera team, incorrect amp-ordering was assumed for data out of the eTraveler. The correct order should be C10,...,17,07,...,00. This is now fixed.

The PR only changes the numbers in policy/lsstCam/Rxx.yaml
by re-assigning the read noise, gain, and saturation values to the correct amps.

`bin.src/update_lsstcam_EO_results.py` is a script we used to help with this update. See https://jira.lsstcorp.org/browse/DM-22824


timj:
I don't think the script should be added to bin.src if it's only used rarely. Maybe put it in examples?

timj:
ie is this a script that you will use again? Or is it a script of historical interest?

bxin:
> ie is this a script that you will use again? Or is it a script of historical interest?

The expectation is that it will be reused when we get more EO data from the camera later.
But I am open to putting it anywhere you suggest.
I put it in `bin.src` mostly because I see `rewrite_ts8_qe_files.py` is in there, which I believe is a script for similar purpose. @SimonKrughoff should chime in.

timj:
Ok. I guess the script name is verbose enough that people won't try to run it without understanding what it is. If @SimonKrughoff is fine with it then that's enough for me.

python/lsst/obs/lsst/script/update_lsstcam_EO_results.py
```diff
@@ -0,0 +1,165 @@
...
+import glob
+
+from lsst.obs.lsst.lsstCamMapper import LsstCamMapper
+
+
+def main():
+    mapper = LsstCamMapper()
+    cam = mapper.camera
+
+    parser = argparse.ArgumentParser(description='Update raft description.')
```
timj, line 47: Update raft for LSSTCam alone? The description should say so.

python/lsst/obs/lsst/script/update_lsstcam_EO_results.py
```diff
@@ -0,0 +1,165 @@
...
+import argparse
+import os
+import glob
+
+from lsst.obs.lsst.lsstCamMapper import LsstCamMapper
+
+
+def main():
+    mapper = LsstCamMapper()
+    cam = mapper.camera
```
timj, line 45: If all you want is camera geom please use gen3:

```python
from lsst.obs.lsst import LSSTCam
instrument = LsstCam()
cam = instrument.getCamera()
```

python/lsst/obs/lsst/script/update_lsstcam_EO_results.py
```diff
@@ -0,0 +1,165 @@
...
+    parser.add_argument('picklefile', type=str,
+                        help='Path to pickle file to use')
+    parser.add_argument('raft_name', type=str,
+                        help='Name of raft to update')
+    parser.add_argument('sensor_type', type=str,
+                        help='Either E2V or ITL')
+    parser.add_argument('out_file', type=str,
+                        help='Name of file to write to')
+
+    args = parser.parse_args()
```
timj, line 57: Please take a look at the other scripts in this directory.  The parser definition should be separated out so that the sphinx documentation can be created for each command. See build_argparser.

python/lsst/obs/lsst/script/update_lsstcam_EO_results.py
```diff
@@ -0,0 +1,165 @@
...
+    print(args.raft_name, '----------------')
+    installed_raft = args.raft_name
+    with open(args.picklefile, mode='rb') as fh:
+        raft = pickle.load(fh)
+        noise = pickle.load(fh)
+        ccd_list = pickle.load(fh)
+        fw = pickle.load(fh)
+        gains = pickle.load(fh)
+
+    sensors = ['S00', 'S01', 'S02', 'S10', 'S11', 'S12', 'S20', 'S21', 'S22']
```
timj, line 68: Can we get the sensors and amps directly from the camera object?

python/lsst/obs/lsst/script/update_lsstcam_EO_results.py
```diff
@@ -0,0 +1,165 @@
...
+    with open(args.out_file, 'w') as fh:
+        fh.write(file_tmpl%(installed_raft, args.sensor_type, raft, "\n".join(
+            serialstrs), "\n".join(geomstrs), "\n".join(sensorstrs))+"\n")
+
+    # deal with the QE curves nows, we need to copy them over from ts8 camera
+    # Based on instructions at https://jira.lsstcorp.org/browse/DM-22824
+
+    # Make directories for the curves to go into:
+    for i in range(len(sensors)):
+        my_folder = f"{os.environ['OBS_LSST_DATA_DIR']}/lsstcam/qe_curves/\
```
timj, line 118: Do not look for the environment variable directly. Instead use:
```python
obsDataDir = instrument.obsDataPackageDir
```

python/lsst/obs/lsst/script/update_lsstcam_EO_results.py
```diff
@@ -0,0 +1,165 @@
...
+    {installed_raft.lower()}_{sensors[i].lower()}"
+        if not os.path.exists(my_folder):
+            # print('creating %s'%my_folder)
+            os.makedirs(my_folder)
+
+    rtmname = raft[-7:]
```
timj, line 124: This needs a comment.  I'd prefer a regex to be sure.

python/lsst/obs/lsst/script/update_lsstcam_EO_results.py
```diff
@@ -0,0 +1,165 @@
...
+    # There are only a few things that need to be updated:
+    # 1. INSTRUME: TS8 --> LSSTCAM
+    # 2. DETECTOR: This needs to be set to the corresponding
+    #    integer in the list above.
+    # 3. CALIB_ID: raftName needs to be changed from the RTM number to
+    #    the actual raft name. detector, ccd and ccdnum need to be set to
+    #    the same thing as DETECTOR.
+
+    for i in range(len(sensors)):
+        srcDir = f"{os.environ['OBS_LSST_DATA_DIR']}/ts8/qe_curve/{rtmname.lower()}_{sensors[i].lower()}"
```
timj, line 136: For filename stuff you really should be using os.path.join (and also not reading the environ again)

python/lsst/obs/lsst/script/update_lsstcam_EO_results.py
```diff
@@ -0,0 +1,165 @@
...
+        destDir = f"{os.environ['OBS_LSST_DATA_DIR']}/lsstcam/qe_curve/\
+{installed_raft.lower()}_{sensors[i].lower()}"
+        srcList = glob.glob('%s/*'%srcDir)
+        assert len(srcList) == 1
+        srcName = srcList[0]
+        destName = os.path.join(destDir, os.path.split(srcName)[-1])
+        rid = open(srcName, 'r')
+        wid = open(destName, 'w')
```
timj, line 144: Use context manager here to ensure the files are closed out.

python/lsst/obs/lsst/script/update_lsstcam_EO_results.py
```diff
@@ -0,0 +1,165 @@
...
+        detectorId = cam[f'{installed_raft}_{sensors[i]}'].getId()
+        for line in rid:
+            if 'TS8' in line:
+                wid.write('# - {INSTRUME: LSSTCAM}\n')
```
timj, line 148: The name is LSSTCam not LSSTCAM.

python/lsst/obs/lsst/script/update_lsstcam_EO_results.py
```diff
@@ -0,0 +1,165 @@
...
+            elif 'DETECTOR' in line:
+                wid.write('# - {DETECTOR: %d}\n'%detectorId)
```
timj, line 150: Why aren't you using `print` here? (and f-strings).

SimonKrughoff:
@timj I have removed the script for now and scheduled a ticket to deal with it later.

PR #244, opened 2020-08-18
DM-26327: Flip LATISS wcs in X for Gen3

timj:


python/lsst/obs/lsst/rawFormatter.py
```diff
@@ -123,6 +123,7 @@ class LatissRawFormatter(LsstCamRawFormatter):
     translatorClass = LatissTranslator
     _instrument = Latiss
     filterDefinitions = Latiss.filterDefinitions
+    wcsFlipX = True
```
kfindeisen, line 126: Should this have a comment analogous to the one in DECam?

PR #247, opened 2020-08-31
DM-26393: Add Object.yaml specification

wmwv:
to transform multiband outputs to the parquet Object Table

### DIFF
--- a/config/imsim/transformObjectCatalog.py
+++ b/config/imsim/transformObjectCatalog.py
@@ -1,0 +1,8 @@
+import os
+from lsst.utils import getPackageDir
+
+# Always produce these output bands in the Object Tables,
+# regardless which bands were processed
+config.outputBands = ["u", "g", "r", "i", "z", "y"]
+
+config.functorFile = os.path.join(getPackageDir('obs_lsst'), 'policy', 'imsim', 'Object.yaml')
--- a/policy/imsim/Object.yaml
+++ b/policy/imsim/Object.yaml
@@ -1,0 +1,424 @@
+funcs:
+    objectId: # the index of deepCoadd_obj IS the objectId
+        functor: Index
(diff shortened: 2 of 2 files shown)


config/imsim/transformObjectCatalog.py
```diff
@@ -0,0 +1,5 @@
+import os
+from lsst.utils import getPackageDir
+
+config.filterMap = {band: band for band in 'ugrizy'}
```
timj, line 4: Does this need to know about the new filter names for imsim?

policy/imsim/Object.yaml
```diff
@@ -0,0 +1,423 @@
...
+    coord_dec:
+        # reference position for merged "ref" cat. Required by db. Not in DPDD
+        # coord_dec because "dec" is reserved in most SQL DBs
+        functor: DecColumn
+        dataset: ref
+    Ra:
+        functor: CoordColumn
+        args: coord_ra
+        dataset: meas
+    Decl:
```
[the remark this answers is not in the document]
yalsayyad: >  Is there something we're using where this is a problem?

qserv/mysql. 

I don't know what to say. I put this ticket on the shelf hoping an idea would appear for giving you an ra/dec here, but I think the best I can offer is a `ra`/`decl` with the same values `coord_ra`/`coord_dec`.

yalsayyad (commented):
response

config/imsim/transformObjectCatalog.py
timj: Ah sorry. I saw "filterMap" and thought this was talking about filters. I missed the `band` on the right. Is this really `bandMap` ?

Note that the physical filter for imsim did change recently to include the sims throughputs version number.

PR #248, opened 2020-08-31
DM-26550: Add support for ObservationInfo.observation_reason

timj:
This allows TESTTYPE handling to be unified.

Requires lsst/astro_metadata_translator#40


config/latiss/ingest.py
```diff
@@ -24,8 +24,7 @@
 
 config.parse.retarget(LatissParseTask)
 
-del config.parse.translation['testType']
-
+del config.parse.translators['testType']
```
RobertLuptonTheGood, line 27: Do we need this white-space only change?

RobertLuptonTheGood (approved):
I don't think that the test `observation_reason`s are very realistic, but as long as no-one cargo cults them that's OK.

config/latiss/ingest.py
timj: It's not a whitespace only change.

timj:
I don't think I can do anything other than believe the value of the TESTTYPE header can I? Garbage in and garbage out. If you think that TESTTYPE=bias could be translated to something else then I'm happy to tweak the translation here.

RobertLuptonTheGood:
Tim explained that the test code was using real `TESTTYPE` values, so my comment is wrong.   Good to merge

PR #249, opened 2020-09-03
DM-26599: Updated raft and sensor positions to be exactly as in LCA-13381

SimonKrughoff:


SimonKrughoff (commented):
This looks good to me.  It has passed CI.

PR #252, opened 2020-09-18
DM-26811: Rename LSST-ImSim to LSSTCam-imSim (and phosim now LSSTCam-PhoSim)

timj:
This makes it clear that these are simulations of LSSTCam and also follows a request from @cwwalter to use "imSim" and not "ImSim".


cwwalter (approved):
This looks good to me. I checked against the file list for completeness and looked for any missed substitutions.  Thanks Tim.

PR #255, opened 2020-09-25
DM-26943: Use AIRMASS or AMSTART headers for imsim if they exist

timj:


python/lsst/obs/lsst/translators/imsim.py
```diff
@@ -113,3 +116,26 @@ def to_physical_filter(self):
...
+            return super().to_altaz_begin()
+
+        location = self.to_location()
+        ha = Angle(self._header["HASTART"], unit=u.deg)
+
+        # For speed over accuracy, assume this is apparent Dec not ICRS
+        dec = Angle(self._header["DECTEL"], unit=u.deg)
+
+        # Use erfa directly
+        import astropy._erfa as erfa
```
ehneilsen, line 137: The import should maybe be moved to the top of the file?

ehneilsen (approved):
I would expect the _erfa import to be at the head of the file with the rest of the imports. Is there a reason it is embedded in the code?
Also, it looks like astropy._erfa has now been deprecated by the astropy team in favor of using a separate standalone pyerfa package. It doesn't look like this is available yet in the version of astropy used here, though.

timj:
Thanks. Yes, I'll try to support pyerfa (I found out about that from the astropy slack over the weekend). I'll also probably move the core hour angle to azel code into astro_metadata_translator since it's generic.

ehneilsen (approved):
Looks good; hopefully this will speed things up considerably.

PR #258, opened 2020-10-09
DM-24355: Make Instrument registration idempotent.

TallJimbo:


python/lsst/obs/lsst/_instrument.py
```diff
@@ -115,18 +115,21 @@ def register(self, registry):
...
+                {
+                    "name": self.getName(),
+                    "detector_max": self.translatorClass.DETECTOR_MAX,
+                    "visit_max": obsMax,
+                    "exposure_max": obsMax,
+                    "class_name": getFullTypeName(self),
+                }
+            )
+            for detector in self.getCamera():
+                registry.syncDimensionData("detector", self.extractDetectorRecord(detector))
```
timj, line 130: Did the bulk insert version disappear or something? Previously it built all the records and did one insert, here it does a separate sync for each detector.

TallJimbo: Correct.  `syncDimensionData` doesn't do bulk.  It probably could be made to now with enough effort, but that's a bigger change than I wanted to make on this ticket, and I think for an operation as rare as instrument registration ought to be, a few hundred single-row insert operations is not a big problem.  I figured being able to add new detectors (and check for consistency with all existing ones) was more valuable than performance in this context.

A no-op sync is not free, though, and I've tried to document that wherever I could.  If people throw calls to `Instrument.register` and the like everywhere (especially `TestCase.setUp`) we will regret it.

PR #259, opened 2020-10-13
DM-27162: Reorganize filter determination for LATISS and LSSTCam

timj:
* Share code for determining the main filter from the FILTER
  header.
* LSSTCam now only adds FILTER2 value if it is not empty.
* Always use UNKNOWN when the filter is not known


python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -664,3 +664,32 @@ def to_exposure_group(self):
...
+        filter : `str`
+            The contents of the ``FILTER`` header with some appropriate
+            defaulting.
+        """
+
+        if self.is_key_ok("FILTER"):
+            physical_filter = self._header["FILTER"]
+            self._used_these_cards("FILTER")
+
+            if physical_filter.lower().startswith("empty"):
```
RobertLuptonTheGood, line 682: Is this a good idea?  If the physical filter is `empty_timj` this'll replace it with `EMPTY`

python/lsst/obs/lsst/translators/lsstCam.py
```diff
@@ -103,3 +103,28 @@ def can_translate(cls, header, filename=None):
...
+            headers joined by a "~" if FILTER2 is set and not empty.
+            Returns "UNKNOWN" if no filter is declared.
+        """
+        physical_filter = self._determine_primary_filter()
+
+        filter2 = None
+        if self.is_key_ok("FILTER2"):
+            self._used_these_cards("FILTER2")
+            filter2 = self._header["FILTER2"]
+            if filter2.lower().startswith("empty"):
```
RobertLuptonTheGood, line 124: Again, I'm nervous about using {{startwith("empty")}}.  I'd prefer to check for a more explicit pattern, e.g. {{re.search(r"^empty[0-9]?$", filter2.lower())}}

RobertLuptonTheGood (commented):
Looks good, except that I'm worried about over-wide matching for empty filter names

python/lsst/obs/lsst/translators/lsst.py
timj: This code came over from LATISS because they have empty_1 and empty_2 but don't care what comes after the underscore. If the convention is that you can have "empty_something_not_empty" then a regex seems to be important but then I start to think that we shouldn't be calling things empty if they aren't empty. Do you see a use case where a filter name could start with empty but absolutely not want to be reported as empty?

timj: I changed to a regex. Jenkins is running now.

PR #260, opened 2020-10-19
DM-27180: NaNs in measurePhotonTransferCurve.py causing failures

plazas:


mfisherlevine (approved):
Definitely a win here 👍

PR #262, opened 2020-10-22
DM-27158: constructDark.py fails on LSSTCam with 'Too many CR pixels'

czwa:
Update R41_S11_C14's readnoise to 25.0, based on BOT data.

This update allows the regular constructDark.py processing to succeed
on this amplifier, as the elevated read noise prevents the cosmic ray
rejection from identifying everything as a cosmic ray.


plazas (approved):
The change itself is trivial, and it looks good if it helps {{constructDark}}. And I also see that you outline, in the Jira ticket, an iterative process that is self-contained within the DM code to deal with this via {{doEmpiricalNoise}}. The more general question is, given that we have gain, full-well capacity at saturation, and noise values from the latest BOT runs (both using eoTtest and DM tools), can't we just update all these YAML files with those more realistic values? Maybe this is a more general discussion.

PR #264, opened 2020-10-27
DM-27339: BOT has unrecognised physical_filters

RobertLuptonTheGood:


RobertLuptonTheGood:
We'll solve this problem by patching the header

PR #266, opened 2020-10-28
Fix header when FILTER2 is set but FILTER isn't; DM-27339

RobertLuptonTheGood:


python/lsst/obs/lsst/translators/lsstCam.py
```diff
@@ -79,6 +79,22 @@ class LsstCamTranslator(LsstBaseTranslator):
...
     cameraPolicyFile = "policy/lsstCam.yaml"
 
+    @classmethod
+    def fix_header(cls, header):
+        """See https://astro-metadata-translator.lsst.io/py-api/astro_metadata_translator.FitsTranslator.html#astro_metadata_translator.FitsTranslator.fix_header"""  # noqa: E501, W505
+
+        if "FILTER" not in header:
+            obsid = header.get("OBSID", "unknown")
+
+            log.warn("%s: No FILTER but FILTER2=\"%s\"", obsid, header.get("FILTER2", "(missing)"))
```
mfisherlevine, line 89: Feels odd to always warn here as if FILTER2 was found, but sometimes it won't be. I think either this should be phrased as a "No FILTER" warning with a supplemental "and FILTER2 was X" warning, or the whole thing put inside an `if` for when FILTER2 _is_ found.

mfisherlevine: If the latter, then you could also add a warning (or perhaps INFO actually) about blanking out the FILTER2 value too.

RobertLuptonTheGood: Hmm, I agree.  Changed to something like:
```       
 if "FILTER" not in header and header.get("FILTER2") is not None:
        log.warn("%s: No FILTER key found but FILTER2=\"%s\" (removed)",
                       header.get("OBSID", "unknown"), header["FILTER2"])
        header["FILTER2"] = None
            
        return True
```
(except that I added the CCD identifier, e.g. `R11_S20`, too)

PR #267, opened 2020-10-30
DM-26696: Support new fix_header API

timj:
Depends on lsst/astro_metadata_translator#43

### DIFF
--- a/python/lsst/obs/lsst/translators/latiss.py
+++ b/python/lsst/obs/lsst/translators/latiss.py
@@ -167,13 +167,21 @@ class LatissTranslator(LsstBaseTranslator):
         return False
 
     @classmethod
-    def fix_header(cls, header):
+    def fix_header(cls, header, instrument, obsid, filename=None):
         """Fix an incorrect LATISS header.
 
         Parameters
         ----------
         header : `dict`
             The header to update.  Updates are in place.
+        instrument : `str`
+            The name of the instrument.
+        obsid : `str`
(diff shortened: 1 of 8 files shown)


python/lsst/obs/lsst/translators/latiss.py
```diff
@@ -196,17 +204,21 @@ def fix_header(cls, header):
...
 
         Corrections are reported as debug level log messages.
+
+        See `~astro_metadata_translator.fix_header` for details of the general
+        process.
         """
         modified = False
 
+        # Prefer filename to obsid for log messages
+        log_label = filename or obsid
```
RobertLuptonTheGood, line 214: Is this pattern documented?  I might well have put both into the log message (I think your choice is fine)

RobertLuptonTheGood (approved):
OK

python/lsst/obs/lsst/translators/latiss.py
timj: It's not documented because it's currently a decision specifically for this method. We could have a class method that is given the file and obsid and returns a consistent string.  You are allowed to put both in but you have to handle the case where there is no file known so we could instead do something like

```python
log_label = f"{filename}:{obsid}" if filename else obsid
```
I'm fine either way.

timj: Following your comment above I have standardized this in a single method and switched all logging to use it. I did add the filename and obsid if filename is defined.

PR #268, opened 2020-11-02
DM-25424: Convert Defect to IsrCalib

czwa:


config/findDefects.py
```diff
@@ -1,23 +1,4 @@
...
-config.isrForDarks.doFlat = False  # don't flatfield darks
-
-# none of the calibration products exists to perform these yet
-for config in [config.isrForDarks, config.isrForFlats]:
-    config.doCrosstalk = False
-    config.doUseOpticsTransmission = False
-    config.doUseFilterTransmission = False
-    config.doUseSensorTransmission = False
-    config.doUseAtmosphereTransmission = False
-    config.doAttachTransmissionCurve = False
```
mfisherlevine, line 23: Some of this looks outdated, but some looks like like it should stay. Did this move somewhere else (_e.g._ to the `setDefaults()` which is arguably where it should always have been) or is it just gone?

config/latiss/findDefects.py
```diff
@@ -19,6 +19,3 @@
 # You should have received a copy of the LSST License Statement and
 # the GNU General Public License along with this program.  If not,
 # see <http://www.lsstcorp.org/LegalNotices/>.
-
-# reassert upstream default as main LSST config sets assertSameRun=True
-config.assertSameRun = False
```
mfisherlevine, line 24: Same might apply to this, depending on the answer to the above.

config/findDefects.py
czwa: This moved to the `cp_pipe/config/defectIsr.py`, which is the gen2 way to pre-process exposures for input to `FindDefectTask`.  In gen3, this is handled by the pipeline definition files.

config/latiss/findDefects.py
czwa: As above.  In addition, as this removes all the code, I'm removing this file entirely.

PR #269, opened 2020-11-04
DM-27281: comcam images from NTS will not ingest in gen3 butler

RobertLuptonTheGood:


python/lsst/obs/lsst/filters.py
```diff
@@ -198,7 +201,7 @@ def addFilter(filter_dict, band, physical_filter, lambdaEff=0.0):
 # LATISS filters include a grating in the name so we need to construct
 # filters for each combination of filter+grating.
 _latiss_filters = (
-    FilterDefinition(physical_filter="NONE",
+    FilterDefinition(physical_filter="EMPTY",
```
timj, line 204: We say "empty" for the physical filter definition for LsstCam above so I think it should be "empty" here as well.

python/lsst/obs/lsst/translators/comCam.py
```diff
@@ -100,6 +104,19 @@ def fix_header(cls, header, instrument, obsid, filename=None):
...
+        if physical_filter in (None, "r", ""):
+            # Create a translator since we need the date
+            translator = cls(header)
+            date = translator.to_datetime_begin()
+            if date > COMCAM_TO_CHILE_DATE:
+                header["FILTER"] = "empty"
+            else:
+                header["FILTER"] = "r_03"  # it's currently 'r', which is a band not a physical_filter
+
+            log.warn("%s: replaced FILTER \"%s\" with \"%s\"", obsid, physical_filter, header["FILTER"])
```
timj, line 117: This has to be `log.warning` (`log.warn` results in a deprecation warning).   Also, you can now switch that `obsid` to `log_label` to be consistent.

timj (approved):
A couple of fixes required but otherwise good.

python/lsst/obs/lsst/translators/comCam.py
timj: It still uses obsid and not log_label so the filename is not reported.

PR #271, opened 2020-11-20
DM-27670: Add filterMap config for imSim processing

yalsayyad:
per filter definitions as of DM-26138.


config/imsim/filterMap.py
```diff
@@ -0,0 +1 @@
+config.filterMap = {f'{_}_sim_1.4': f'lsst_{_}_smeared' for _ in 'ugrizy'}
```
taranu, line 1: I suggest changing the `_` variable to `band` for readability.

taranu (changes requested):

PR #274, opened 2020-12-04
DM-27863: added controller == "P" option to compute_exposure_id

suberlak:


python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -398,8 +398,12 @@ def compute_exposure_id(dayobs, seqnum, controller=None):
                 # Add 1000 to the year component
                 dayobs = int(dayobs)
                 dayobs += 1000_00_00
+            elif controller == "P":
```
ktlim, line 401: Shouldn't this be "S"?  "P" means "playback" according to https://confluence.lsstcorp.org/display/SYSENG/Image+naming+and+visit+definition+conventions

timj: The problem we have is that "S" is far too generic. Imsim and Phosim can't clash so you need different controllers for each and "S" may well mean that the data are simulated by the DAQ. Have we ever taken data in playback mode? (I'm not sure what that means because if you are playing back previous data why would it be a different controller code).

Maybe we could use H for "pHosim"? It doesn't really matter so long as we pick something unique for phosim.

ktlim: The DAQ doesn't simulate data; it simulates data-taking.  It can either do so with the original metadata ("playback") or with new metadata, as if these are new images ("simulation").

Are you saying we need to distinguish between both of these and images that have never been through a DAQ? It's not clear to me why the images are any different from the pipeline's point of view.

If the point is that we may have many image generators and that they're all going into the same repo with the same instrument, I'm not sure hard-coded letters are going to be sufficient.  Wouldn't my Phosim run and someone else's Phosim run potentially clash?

timj: I'm saying I need to guarantee unique OBSID values for phosim data if they are starting to generate phosim data with LSSTCam headers. The easiest way to guarantee that uniqueness was to allocate phosim a controller code -- this is because from a pipeline perspective the files will be LSSTCam data. Currently phosim is a special instrument LSSTCamPhoSim so obsid clashes are impossible.

suberlak:
As requested, I changed "P" to "H", can this update be merged?

ktlim:
The process is for you to put DM-27863 into "In Review" state and choose a code reviewer.  Once it has been reviewed and Jenkins CI runs have passed, you may merge.

python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -398,8 +398,12 @@ def compute_exposure_id(dayobs, seqnum, controller=None):
                 # Add 1000 to the year component
                 dayobs = int(dayobs)
                 dayobs += 1000_00_00
+            elif controller == "H":
```
ktlim, line 401: Please add a comment to indicate what this is for, either here, or perhaps even better under the "Camera control" comment above.

PR #275, opened 2020-12-16
DM-27863: Allow controller = 'H' for exposure_id calculation

suberlak:


timj:
I don't think we should be putting a phosim lsst cam test file in data/input because those are meant to be gen2 butler repos. I think what should be happening is that the FITS header should be written to tests/headers and the test should be in test_translator.py -- that would test the header translation is working properly and wouldn't rely on old gen2 infrastructure.

Also, the INSTRUME header should be LSSTCam.

ktlim (approved):

timj:
@suberlak in the future please rebase your branch before merging and do not use GitHub to update it -- if you do that it merges master to the ticket branch and that is not DM convention.

PR #277, opened 2020-12-22
DM-28046: Updated PhosimEimgParseTask to work with LSSTCam-style Phosim data

suberlak:
Updated PhosimEimgParseTask and removed the old test


timj:
Please remove that "merge branch master into ticket" commit and do a proper rebase. We never use the "update branch" button.

tests/test_parsetask.py
```diff
@@ -314,33 +314,6 @@ def test_parsetask_phosim_translator(self):
                      )
         self.assertParseCompare(DATADIR, CONFIGDIR, "phosim", PhosimParseTask, test_data)
 
-    def test_parsetask_phosim_eimg_translator(self):
```
timj, line 317: Rather than removing this test, a corrected eimage file (properly stripped down to make it small) should replace this test file (which should be deleted).

python/lsst/obs/lsst/phosim.py
```diff
@@ -59,7 +59,7 @@ def translate_controller(self, md):
         return "S"
 
 
-class PhosimEimgParseTask(PhosimParseTask):
+class PhosimEimgParseTask(LsstCamParseTask):
```
timj, line 62: Are you okay with breaking all existing eimage files?

Shouldn't there be a new `LsstCamEimgParseTask` instead?

timj (approved):
Looks fine although I'm worried about old data now being broken and that there is no test that new data header parsing works.

python/lsst/obs/lsst/phosim.py
suberlak: As far as I can tell, the only ingest config file that uses `PhosimEimgParseTask` is https://github.com/lsst/obs_lsst/blob/master/config/phosim/ingestEimg.py  .  Changing `PhosimEimgParseTask` would break the ingest of   e-images repackaged with `PhosimMapper`,  but this is par for the course of  updating the repackager to use `LsstCamMapper`.  However,  instead of changing `PhosimEimgParseTask` and associated  unit test, I can change `ingestEimg.py`, to point to a new `LsstCamEimgParseTask` (that way no changes are needed to ts_wep in places where e-images are ingested with `ingestEimg.py` config file ). This is short-lived since soon we will update ts_wep  to use Gen3 repos, but that would be next in line.

suberlak:
I followed a suggestion to keep `phosim.py` and `PhosimEimgParseTask` unchanged, and instead make a new `LsstCamEimgParseTask`. I added that to `ingest.py` and a necessary config file  `config/lsstCam/ingestEimg.py` .

timj (approved):
Please rebase (not merging from master) before merging.

timj:
and make sure jenkins passes.

suberlak:
ok, there's no "merge from master" commit,  but jenkins has errors that are unrelated to anything I changed - https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/33657/tests (looks like gen3 tests).  Do I need to merge master into the branch to solve those? If I run `scons` in the ticket branch it runs successfully, that's why I pushed all changes.

timj:
You need to rebase.  If you don't know how to do that search for rebase in the dev guide.  I am specifically saying do not merge master into your branch.

suberlak:
Ok,  thank you!  It got rebased to master, and Jenkins now passes successfully.

PR #278, opened 2020-12-23
DM-28103: Load camera-specific configs for SafeClipAssemble subtasks

isullivan:


timj:
@isullivan can we close this PR? It looks like the ticket was completed.

PR #279, opened 2020-12-30
DM-28158: Remove assertion on times that are no longer top far in the future.

TallJimbo:


isullivan (approved):
Please fix the typo in the commit message ("top" -> "too"?), otherwise this looks fine.

I'm a little worried this might trip us up again in the future, when the ImSim data might instead be too far in the past. Could you instead catch this specific type of error with a try/except block, or would that defeat the test?

tests/test_translator.py
```diff
@@ -681,13 +681,7 @@ def test_imsim_translator(self):
...
         for filename, expected in test_data:
             with self.subTest(f"Testing {filename}"):
-                # ImSim data are in the future and Astropy complains
-                # about astrometry errors.
-                if expected["observation_type"] == "science":
-                    with self.assertWarns(astropy.utils.exceptions.AstropyWarning):
-                        self.assertObservationInfoFromYaml(filename, dir=self.datadir, **expected)
-                else:
-                    self.assertObservationInfoFromYaml(filename, dir=self.datadir, **expected)
+                self.assertObservationInfoFromYaml(filename, dir=self.datadir, **expected)
```
timj, line 684: The only comment is that this test was mainly there to *hide* the warning from the test output so that people don't get confused when they run pytest and see some unexpected warnings turning up. I assume that is not a problem here because now we've hit a time where imsim will never trigger more warnings.

TallJimbo:
> I'm a little worried this might trip us up again in the future, when the ImSim data might instead be too far in the past. Could you instead catch this specific type of error with a try/except block, or would that defeat the test?

I don't _think_ "too far in the past" is a thing; I think the logic astropy is worried about projecting into the future is based on some kind of table for which historical data is not a problem (at least in the modern era).

PR #280, opened 2021-01-02
DM-27170: Deprecate and phase out afw::image::Filter

parejkoj:


python/lsst/obs/lsst/lsstCamMapper.py
```diff
@@ -464,9 +464,9 @@ def std_raw(self, item, dataId, filter=True):
                             obsInfo.observation_id, obsInfo.detector_unique_name,
                             unknownName)
 
-                filt = afwImage.Filter(unknownName)
+                filt = afwImage.FilterLabel(physical=unknownName)
```
kfindeisen, line 467: Now that we aren't forced to include a filter in each `Exposure`, I don't think we should keep this behavior. It will be much clearer to downstream users if the `Exposure` has no filter at all rather than one with a magic value.

tests/test_ingest.py
```diff
@@ -40,6 +40,7 @@ class LatissIngestTestCase(IngestTestBase, lsst.utils.tests.TestCase):
     ingestDir = TESTDIR
     file = os.path.join(DATAROOT, "latiss", "raw", "2018-09-20", "3018092000065-det000.fits")
     dataIds = [dict(instrument="LATISS", exposure=3018092000065, detector=0)]
+    filterLabel = lsst.afw.image.FilterLabel(physical="unknown~unknown")
```
kfindeisen, line 43: Are these test cases trying to reproduce the old Filter behavior (where you'd always have a default filter called "\_unknown\_")?

kfindeisen (commented):
I'm worried these changes are a bit too wedded to the old `Exposure`, which must have a `Filter` even if no filter information is available. Why do three of the four test cases say "unknown"?

tests/test_ingest.py
timj: `unknown~unknown` is the actual name we use for the LATISS physical filter if there is a bug in the old data header and we don't know what the filter was.

parejkoj:
Answering your question re: "unknown" in the tests: that's what the filters in those files are called. I'm assuming this is very early data for those cameras and they came from before various headers were standardized.

python/lsst/obs/lsst/lsstCamMapper.py
parejkoj: Looking at the logic some more here, the new code will not raise `LookupError` at all, since that was a feature of the `Filter` singleton. I've now removed that entire block.

@timj : could you tell me if you're aware of anything that depends on the previous behavior here?

timj: The new logic there seems reasonable.

kfindeisen (approved):

PR #283, opened 2021-02-02
DM-28612: Enable tracking ra/dec calculation for LSSTCam

timj:


python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -637,7 +637,7 @@ def to_tracking_radec(self):
         if self._header["DATE-OBS"] == self._header["DATE"]:
             # A fixed up date -- use AZEL as source of truth
             altaz = self.to_altaz_begin()
-            radec = astropy.coordinates.SkyCoord(altaz.transform_to(astropy.coordinates.ICRS),
+            radec = astropy.coordinates.SkyCoord(altaz.transform_to(astropy.coordinates.ICRS()),
```
ktlim, line 640: If I'm reading correctly, `transform_to()` can accept a class as well as an instance.  Why the more verbose usage?

python/lsst/obs/lsst/translators/lsstsim.py
```diff
@@ -81,9 +81,9 @@ def to_altaz_begin(self):
             # Derive from RADec in absence of any other information
             radec = self.to_tracking_radec()
             if radec is not None:
-                # This triggers warnings because of the future dates
+                # This can trigger warnings because of the future dates
                 with warnings.catch_warnings():
                     warnings.simplefilter("ignore", category=astropy.utils.exceptions.AstropyWarning)
-                    altaz = radec.transform_to(AltAz)
+                    altaz = radec.transform_to(AltAz())
```
ktlim, line 87: Same here about class vs. instance.

ktlim (approved):

python/lsst/obs/lsst/translators/lsst.py
timj: That's what I thought too but in astropy 4.2 the code issues a deprecation warning if a class is used.

PR #286, opened 2021-02-08
Tickets/dm 28553

RobertLuptonTheGood:
Fixed critical error identified by KT.

### DIFF
--- a/policy/lsstCam/R04.yaml
+++ b/policy/lsstCam/R04.yaml
@@ -23,131 +23,163 @@ R04 :
   amplifiers :
     SG0 :
       C10 : {
-        gain : 1.62900,
-        readNoise : 4.210,
+        gain : 1.60000,
+        readNoise : 14.141,
+        saturation : 71014,
       }
       C11 : {
-        gain : 1.79600,
-        readNoise : 3.860,
+        gain : 0.90000,
+        readNoise : 11.537,
+        saturation : 27,
       }
       C12 : {
-        gain : 1.62300,
-        readNoise : 4.270,
+        gain : 1.29162,
+        readNoise : 16.443,
+        saturation : 74749,
       }
       C13 : {
-        gain : 1.75100,
-        readNoise : 3.900,
+        gain : 1.29254,
+        readNoise : 13.853,
+        saturation : 112836,
       }
       C14 : {
-        gain : 1.66500,
-        readNoise : 4.230,
+        gain : 1.33044,
+        readNoise : 13.036,
+        saturation : 69709,
       }
       C15 : {
-        gain : 1.65900,
-        readNoise : 4.180,
+        gain : 1.26197,
+        readNoise : 11.651,
+        saturation : 80218,
       }
       C16 : {
-        gain : 1.64400,
-        readNoise : 4.340,
+        gain : 1.25942,
+        readNoise : 10.710,
+        saturation : 81895,
       }
       C17 : {
-        gain : 1.69700,
-        readNoise : 4.070,
+        gain : 1.22001,
+        readNoise : 11.795,
+        saturation : 68418,
       }
       C07 : {
-        gain : 1.72200,
-        readNoise : 3.950,
+        gain : 1.11891,
+        readNoise : 12.787,
+        saturation : 100078,
       }
       C06 : {
-        gain : 1.64500,
-        readNoise : 4.350,
+        gain : 1.13583,
+        readNoise : 10.164,
+        saturation : 96865,
       }
       C05 : {
-        gain : 1.62100,
-        readNoise : 4.130,
+        gain : 1.11721,
+        readNoise : 10.189,
+        saturation : 95997,
       }
       C04 : {
-        gain : 1.61900,
-        readNoise : 4.240,
+        gain : 1.11698,
+        readNoise : 10.160,
+        saturation : 94047,
       }
       C03 : {
-        gain : 1.69000,
-        readNoise : 4.260,
+        gain : 1.13458,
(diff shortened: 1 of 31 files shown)


policy/lsstCam/R30.yaml
```diff
@@ -53,155 +53,731 @@ R30 :
...
-      C05 : { gain : 0.787388, readNoise : 4.982099, saturation : 0.000000 }
-      C04 : { gain : 0.781836, readNoise : 4.966761, saturation : 0.000000 }
-      C03 : { gain : 0.783005, readNoise : 4.988240, saturation : 0.000000 }
-      C02 : { gain : 0.784215, readNoise : 4.990350, saturation : 0.000000 }
-      C01 : { gain : 0.777559, readNoise : 4.985713, saturation : 0.000000 }
-      C00 : { gain : 0.766732, readNoise : 4.981440, saturation : 0.000000 }
+      C10 : {
+        gain : 1.60000,
+        readNoise : 4.553,
+        saturation : 0,
```
ktlim, line 59: This is the only one that still has a 0 saturation.  Is that correct?

policy/lsstCam/R04.yaml
```diff
@@ -22,36 +22,164 @@ R04 :
...
-      C00 : { gain : 1.711, readNoise : 4.13 }
+      C10 : {
+        gain : 1.60000,
+        readNoise : 14.141,
+        saturation : 71014,
+      }
+      C11 : {
+        gain : 0.90000,
+        readNoise : 11.537,
+        saturation : 27,
```
ktlim, line 33: This is very low but non-zero.  I assume it's correct?

policy/lsstCam/R40.yaml
```diff
@@ -22,36 +22,164 @@ R40 :
...
-      C06 : { gain : 1.726, readNoise : 4.12 }
-      C05 : { gain : 1.753, readNoise : 4.04 }
-      C04 : { gain : 1.713, readNoise : 4.17 }
-      C03 : { gain : 1.716, readNoise : 3.95 }
-      C02 : { gain : 1.671, readNoise : 4.16 }
-      C01 : { gain : 1.654, readNoise : 4.31 }
-      C00 : { gain : 1.702, readNoise : 4.28 }
+      C10 : {
+        gain : 1.00126,
+        readNoise : 123.163,
```
ktlim, line 108: This is very high; I assume it is correct.

policy/lsstCam/R41.yaml
```diff
@@ -53,155 +53,731 @@ R41 :
...
+        saturation : 79504,
+      }
+      C03 : {
+        gain : 1.42779,
+        readNoise : 6.371,
+        saturation : 79570,
+      }
+      C02 : {
+        gain : 1.43896,
+        readNoise : 110.150,
```
ktlim, line 690: This is very high; I assume it is correct.

policy/lsstCam/R44.yaml
```diff
@@ -22,36 +22,164 @@ R44 :
...
+        saturation : 103376,
+      }
+      C03 : {
+        gain : 1.17464,
+        readNoise : 14.212,
+        saturation : 99776,
+      }
+      C02 : {
+        gain : 1.60000,
+        readNoise : 68.338,
```
ktlim, line 92: This is quite high; I assume it is correct.

python/lsst/obs/lsst/script/updateFromEOTest.py
```diff
@@ -0,0 +1,202 @@
...
+import yaml
+
+import lsst.utils
+
+
+def writeRXXFile(fileName, raftName, raftNameData, gains=None, readNoises=None, saturations=None, indent=2):
+
+    dindent = indent
+
+    with open(fileName, "w") as fd:   # can't use `with` because of the sys.stdout workaround
```
ktlim, line 42: Seems strange to say "can't use with" when it's a `with` statement...

ktlim (approved):

python/lsst/obs/lsst/script/updateFromEOTest.py
RobertLuptonTheGood: Drat, I fixed that (but in Princeton).  Fixed and squashed.

bxin:
Here is a comparison before & after this update (just for the read noise):
image

Just looking at the overall distribution (with a color bar ranging 0-18 e-), the change in the overall pattern is bigger than what I would expect. For example, if we look at R34, R24, and the not-so-good amps on R41, R10, R01, and R31.

Are these just coming from different EO runs for the same sensors? What is causing these (large) fluctuations between the runs?

jchiang87:
> Just looking at the overall distribution (with a color bar ranging 0-18 e-), the change in the overall pattern is bigger than what I would expect. For example, if we look at R34, R24, and the not-so-good amps on R41, R10, R01, and R31.

>Are these just coming from different EO runs for the same sensors? What is causing these (large) fluctuations between the runs?

The read noise will depend a lot on the sequencer files used, and in particular, it will depend on the readout time.  Looking at the BOT data since 2020-11-03 (during I&T's "meta" Run 4), the read noise for rafts R34 and R24 are as stable as the other science rafts over the course of 10-16 different runs where the read noise was measured (not all CCDs were on for all of these runs). These runs all used 2s readout sequencer files with some adjustments among the different runs.  So I'd say the noticeable differences between the read noise values in this PR vs those in JSR2020 can likely be attributed to different settings, most likely in the sequencer files used.  The individual noisy amps in R01, R41, and R43 are known problem cases and have read noise that varies a lot from run to run in addition to being generally noisy.

PR #287, opened 2021-02-09
DM-28459: Standardize LSSTCam and LATISS filter band derivations

timj:


python/lsst/obs/lsst/filters.py
```diff
@@ -216,39 +206,51 @@ def addFilter(filter_dict, band, physical_filter, lambdaEff=0.0):
...
 # Include the filters without the grating in case someone wants
 # to retrieve a filter by an actual filter name
 _latiss_filter_and_grating = [f for f in _latiss_filters]
 
 for filter in _latiss_filters:
     for grating in _latiss_gratings:
+        # The diffuser "filter" was never used with gratings
+        # so skip it
+        if filter.physical_filter == "diffuser":
+            continue
```
mfisherlevine, line 239: I think this is likely fine, I just worry that just because that wasn't happened to date, that doesn't mean it _never_ will in future. But I think this is probably fine for now, give we're really trying to keep the numbers down.

mfisherlevine (approved):

python/lsst/obs/lsst/filters.py
timj: If new diffuser data turns up we can sort it out at the time.

PR #291, opened 2021-02-18
DM-24638: Add DRP pipeline specification

yalsayyad:
and add config override files for tasks with
changed DefaultNames in Gen3

### DIFF
--- a/pipelines/DRP.yaml
+++ b/pipelines/DRP.yaml
@@ -1,0 +1,5 @@
+description: DRP pipeline applicable to all cameras in obs_lsst
+imports:
+  - location: $PIPE_TASKS_DIR/pipelines/DRP.yaml
+    exclude:
+      - sourceTable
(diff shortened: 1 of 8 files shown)


pipelines/DRP.yaml
```diff
@@ -0,0 +1,5 @@
+description: DRP pipeline applicable to all cameras in obs_lsst
+imports:
+  - location: $PIPE_TASKS_DIR/pipelines/DRP.yaml
```
parejkoj, line 3: Out of curiosity, what is specific to LSST in this? Couldn't it live in a non-instrument package?

yalsayyad: See convo on obs_subaru. Source table only works on obs_lsst/imsim and obs_subaru until I do DM-28862. Instead of putting the sourceTable tasks in the obs_lsst/imsim and obs_subaru DRP pipelines, @natelust preferred me to add it to pipe_tasks and exclude it everywhere but obs_lsst/imsim and obs_subaru. I don't care. I just want to merge the gen3-fied source table tasks.

PR #292, opened 2021-02-22
Adding better gain/noise numbers for AuxTel and ComCam

craiglagegit:
I propose to update the gain and noise numbers in the yaml files for ComCam and LATISS.  I have much better numbers than what are in the files today.  The AuxTel PTC run is at /project/shared/auxTel/rerun/cslage/PTC_Defect_2021-02-18/, and the ComCam data is at: /project/shared/comCam-CCS/rerun/cslage/PTC_2020-12-29.  I could also do the LSSTCam if people agree, but it is further away from being finalized.


craiglagegit:
Updated the noise numbers using doEmpiricalReadNoise=True in runISR.py, which reduces the noise numbers to half or less of what was calculated using the PTC intercept.

timj (approved):
Test change looks good to me.

PR #293, opened 2021-03-03
DM-29019: Interpret detector_max as exclusive bound on zero-based detector_nums.

TallJimbo:
The naming isn't great, but this is the interpretation that daf_butler and other cameras have assumed, and we need obs_lsst to be the same in order for DimensionPacker to behave consistently with astro_metadata_translator until we can unify that log.

This removes the DETECTOR_MAX override for TS3 because it should just inherit the same value from its base class anyway.

### DIFF
--- a/python/lsst/obs/lsst/translators/lsst.py
+++ b/python/lsst/obs/lsst/translators/lsst.py
@@ -126,11 +126,11 @@ def compute_detector_exposure_id_generic(exposure_id, detector_num, max_num=1000
 
     if detector_num is None:
         raise ValueError("Detector number must be defined.")
-    if detector_num > max_num or detector_num < 0:
-        raise ValueError(f"Detector number out of range 0 <= {detector_num} <= {max_num}")
+    if detector_num >= max_num or detector_num < 0:
+        raise ValueError(f"Detector number out of range 0 <= {detector_num} < {max_num}")
 
     if mode == "concat":
-        npad = len(str(max_num))
+        npad = len(str(max_num - 1))
         return int(f"{exposure_id}{detector_num:0{npad}d}")
     elif mode == "multiply":
         return max_num*exposure_id + detector_num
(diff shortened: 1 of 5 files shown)


python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -126,11 +126,11 @@ def compute_detector_exposure_id_generic(exposure_id, detector_num, max_num=1000
 
     if detector_num is None:
         raise ValueError("Detector number must be defined.")
-    if detector_num > max_num or detector_num < 0:
-        raise ValueError(f"Detector number out of range 0 <= {detector_num} <= {max_num}")
+    if detector_num >= max_num or detector_num < 0:
+        raise ValueError(f"Detector number out of range 0 <= {detector_num} < {max_num}")
 
     if mode == "concat":
```
timj, line 132: If we are worried about butler compatibility I think we should get rid of concat completely and always use the multiply option. This will completely remove all the `-1` shenanigans and make the DETECTOR_MAX declaration (which is used by the `Instrument` classes) directly compatible with butler.

python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -154,9 +154,16 @@ class LsstBaseTranslator(FitsTranslator):
     detectorSerials = None
     """Mapping of detector serial number to raft, number, and name."""
 
-    DETECTOR_MAX = 999
+    DETECTOR_MAX = 1000
     """Maximum number of detectors to use when calculating the
-    detector_exposure_id."""
+    detector_exposure_id.
+
+    Note that because this is the maximum number *of* exposures, for
```
timj, line 161: Do you mean "exposures" here?

timj (changes requested):
I think this is fixing the wrong thing and we need to conform explicitly by removing the concat option completely.

python/lsst/obs/lsst/translators/lsst.py
TallJimbo: Nope, should have been "detectors".  Will fix.

python/lsst/obs/lsst/translators/lsst.py
TallJimbo: Will do.

Just so we're on the same page, that will _not_ get rid of the new `-1` down in `max_detector_exposure_id`, unless you want to declare that to be an exclusive bound, too, and change the tests accordingly.

TallJimbo: Ok, I've done that - and it required that I change the DETECTOR_MAX values for TS8 and UCDCam, in order to keep the ID calculations the same as before (which I assume is more important than keeping the DETECTOR_MAX values the same as before).  More info on the commit messages.  Mind taking another look?

timj (approved):

PR #295, opened 2021-03-12
DM-29187: Do not fix RASTART header for LATISS after 2021-02-12

timj:


RobertLuptonTheGood (approved):
Looks fine.  I'm not qualified to comment on when the time was correct (but put the general question of notifying people of this sort of change on the next CAP meeting agenda)

timj:
Who does know when it was fixed? I probably shouldn't merge it until we know. @menanteau do you know when the 37s offset was fixed in RASTART or is that more of a @tribeiro question?

menanteau:
@timj I think this was related to the `timestamp` from Camera was sent in `UTC` instead of `TAI`. But I can be mistaken.

timj:
I imagine the question is whether anyone cares about the difference in reality. If I declare it to be correct on that date but it was in fact fixed a week later the only thing that really happens is that the WCS correction will have to do more work. If that's the case I'll merge tomorrow for the weekly if no-one else jumps in.

PR #298, opened 2021-03-20
Recognize phosim-looking-like-comcam

suberlak:
A small update to phosim translator, to recognize that if 'INSTRUME' is in the header, it's supposed not to use phosim translator (but ComCam or LsstCam instead)

### DIFF
--- a/python/lsst/obs/lsst/translators/phosim.py
+++ b/python/lsst/obs/lsst/translators/phosim.py
@@ -10,7 +10,7 @@
 
 """Metadata translation code for LSSTCam PhoSim FITS headers"""
 
-__all__ = ("LsstCamPhoSimTranslator", )
+__all__ = ("LsstCamPhoSimTranslator",)
 
 import logging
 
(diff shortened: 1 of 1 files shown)


python/lsst/obs/lsst/translators/phosim.py
```diff
@@ -81,8 +81,11 @@ def can_translate(cls, header, filename=None):
             `True` if the header is recognized by this class. `False`
             otherwise.
         """
-        return cls.can_translate_with_options(header, {"CREATOR": "PHOSIM", "TESTTYPE": "PHOSIM"},
-                                              filename=filename)
+        if "INSTRUME" in header:  # trying to look like real data
```
timj, line 84: Maybe add a proper comment along the lines of:

```python
# Generic PhoSim data does not have an INSTRUME header.
# If an INSTRUME header is present this translator class is not suitable.
```

timj (approved):
Looks right to me.

python/lsst/obs/lsst/translators/phosim.py
suberlak: Thank you, I updated the comment

PR #299, opened 2021-03-22
DM-29266: Allow phosim to merge primary and secondary header for ingest

timj:


kfindeisen (commented):
I have a few reservations about the `read_all_headers` implementation, but they're analogous to those on lsst/astro_metadata_translator#52.

PR #300, opened 2021-03-22
DM-29336: Templates are identical for icExp and icExpBackground in obs_lsst

laurenam:
The policy templates were identical for the icExp and icExpBackground datasets
thus a "put" of one would overwrite the other.  Fixed here by updating the
template of the latter.


mfisherlevine (approved):
Looks great.

PR #302, opened 2021-03-26
DM-29424: Enable boresight translation for comCam

suberlak:
Remove in `LsstComCam` translator  the `boresight_rotation_coord`, `boresight_rotation_angle`,  `boresight_airmass`, `tracking_radec` from `_const_map` so that they can be calculated.  Fix header  if `ROTPA` not present. Update `test_translator.py` adding a test and test file for phosim-looking-like-comcam (`H` controller)


python/lsst/obs/lsst/translators/comCam.py
```diff
@@ -135,4 +131,9 @@ def fix_header(cls, header, instrument, obsid, filename=None):
                 modified = True
                 log.debug("%s: Set LSST_NUM to %s", log_label, header["LSST_NUM"])
 
+        if "ROTPA" not in header:
+            header["ROTPA"] = 0.0
+            log.warning("Missing ROTPA in header - replacing with None")
```
timj, line 136: You are replacing with 0.0 now.

timj (approved):

python/lsst/obs/lsst/translators/comCam.py
suberlak: Yes, corrected!

PR #307, opened 2021-04-21
DM-29699: CalibCombineConnections changes its quantum dimensions at construction

czwa:
Use CalibCombineByFilterTask where appropriate.

### DIFF
--- a/pipelines/imsim/cpFlat.yaml
+++ b/pipelines/imsim/cpFlat.yaml
@@ -33,13 +33,12 @@ tasks:
       connections.inputMDs: 'flatStats'
       connections.outputScales: 'cpFlatNormScales'
   cpFlatCombine:
-    class: lsst.cp.pipe.cpCombine.CalibCombineTask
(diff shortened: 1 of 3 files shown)


mrawls:
In the spirit of RFC-775, please migrate these pipelines out of `obs_*` and put them in `cp_pipe/pipelines/[camera]` instead. I for one have been sending folks example DECam `makeFlat.yaml` pipelines via Slack DM because I'm bad at following my own guidelines... but in the (very near!) future I will be thrilled to point them to `cp_pipe/pipelines`. Especially now that I know there are breaking changes happening.

czwa:
I've created a new ticket for the migration of camera specific pipelines: DM-29981.  There are pipelines in `obs_subaru` as well (which inherit from the `cp_pipe` versions, and so didn't need edits here), and I'd rather do the migration at once.

mrawls (approved):
Approving with the understanding these pipelines will ultimately land in cp_pipe, not obs_lsst.

PR #313, opened 2021-05-19
DM-29370: support for new detector component and 'amp' parameter in raw formatters

TallJimbo:


python/lsst/obs/lsst/assembly.py
```diff
@@ -105,6 +109,12 @@ def fixAmpGeometry(inAmp, bbox, metadata, logCmd=None):
         def logCmd(*args):
             return
 
+    # check that the book-keeping worked and we got the correct EXTNAME
+    extname = metadata.get("EXTNAME")
+    predictedExtname = f"Segment{inAmp.getName()[1:]}"
+    if extname is not None and predictedExtname != extname:
+        logger.warning('expected to see EXTNAME == "%s", but saw "%s"', predictedExtname, extname)
```
czwa, line 116: I know this is just a migration of this check, but if I do `butler.get('raw', {'amp': 'R12_S00'})`, isn't it an error if I receive something else?  I can see in the full-detector case that getting the wrong thing might be fine, because each amp will mosaic into the right place eventually, but I'm not sure that's the case for single-amp reads.

czwa (approved):
I'm slightly concerned that we only warn if we read a extension with one when we expect another, but I don't know how much that warning triggers (I can imagine it only happens for very old data).

python/lsst/obs/lsst/assembly.py
timj: `amp` doesn't take a detector name does it? I haven't looked at the code but shouldn't the amp parameter be an integer?

czwa: From the `obs_base` ticket branch, `formatters/fitsExposure.py` line 458 seems to allow an integer or a string,

TallJimbo: I assume this is a guard against incorrect headers or something, but I don't actually know; the check/warning was there before, and all I did here was move it around and make it actually work.

PR #314, opened 2021-05-26
DM-30342: Add detect flags to object tables

fred3m:


policy/imsim/Object.yaml
```diff
@@ -318,7 +318,13 @@ refFlags:
     - detect_isPatchInner
     - detect_isPrimary
     - detect_isTractInner
+    - detect_isIsolated
+    - detect_fromBlend
+    - detect_isDeblendedSource
+    - detect_isDeblendedModelSource
+    - merge_peak_sky
     - calib_psf_used
+    - merge_measurement_band
```
jeffcarlin, line 327: From what I can tell, this should simply be `merge_measurement`. Then add a renaming rule at the bottom like `['merge_', 'Merge_']`. (The band will be appended at the beginning to create, e.g., `zMerge_measurement`.)

policy/imsim/Source.yaml
```diff
@@ -252,6 +252,12 @@ flags:
    - ext_shapeHSM_HsmShapeRegauss_flag_parent_source
    - sky_source
    - detect_isPrimary
+   - detect_isIsolated
+   - detect_fromBlend
+   - detect_isDeblendedSource
+   - detect_isDeblendedModelSource
+   - merge_peak_sky
+   - merge_measurement_band
```
jeffcarlin, line 260: See comment above. Note also that the same renaming rule proposed above is needed for `merge_peak_sky`, too.

jeffcarlin (approved):
After the suggested updates, I think it's good to go.

policy/imsim/Object.yaml
fred3m: But where should it go? The comment in Object.yaml says `# refFlags are columns taken without translation from the ref table`, so it sounds like the band isn't passed through them. And the flags section says `# flags are columns taken and exploded per-band from the meas tables`, but I don't think that these columns exist in the meas tables.

fred3m: I think that I'm just going to call `merge_measurement` out of scope and assign that to a new ticket.

jeffcarlin: I'm OK with calling `merge_measurement` out of scope for this one. But I think you will still need the renaming rule for `merge_` --> `Merge_`?

fred3m: Yup, that change is about to be pushed.

PR #337, opened 2021-08-17
DM-31412: Add config/imsim/charImage.py to obs_lsst

leeskelvin:


laurenam:
Wouldn’t hurt to say in the commit message why this is necessary (i.e. it’s a stub until gen2 deprecation).

laurenam (approved):

PR #338, opened 2021-08-18
DM-31445: Enable GAaP PSF photometry before apCorr

arunkannawadi:
`doPsfPhotometry` must be set to `True` prior to registering the plugin for
aperture corrections.

### DIFF
--- a/config/forcedPhotCoadd.py
+++ b/config/forcedPhotCoadd.py
@@ -35,6 +35,9 @@ config.measurement.slots.gaussianFlux = None
 config.catalogCalculation.plugins.names = ["base_ClassificationExtendedness"]
 config.measurement.slots.psfFlux = "base_PsfFlux"
 
+# Enable PSF photometry after PSF-Gaussianization in the `ext_gaap_GaapFlux` plugin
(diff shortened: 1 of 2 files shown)


config/forcedPhotCoadd.py
```diff
@@ -35,6 +35,9 @@
 config.catalogCalculation.plugins.names = ["base_ClassificationExtendedness"]
 config.measurement.slots.psfFlux = "base_PsfFlux"
 
+# Enable PSF photometry after PSF-Gaussianization in the `ext_gaap_GaapFlux` plugin
+config.measurement.plugins["ext_gaap_GaapFlux"].doPsfPhotometry = True
+
```
erykoff, line 40: I'm just curious why this reordering was necessary?  And if there's some missing documentation, can that be ticketed?

erykoff (approved):

config/forcedPhotCoadd.py
arunkannawadi: I wanted this to appear before `doUndeblended` function call on the GAaP plugin. Basically, this has to be set to True before the `getAllGaapResultNames()` method is called. It may not have any effect whatsoever, but wanted to be doubly safe.

PR #340, opened 2021-08-30
DM-31580: Move consolidateDiaSourceTable to step6

hsinfang:


pipelines/imsim/DRP.yaml
```diff
@@ -109,22 +109,25 @@ subsets:
...
         - drpAssociation
         - drpDiaCalculation
         description: >
             Tasks that can be run together, but only after the 'step1', 'step2',
             'step3', and 'step4' subsets
 
-            These DIA aggregation steps include both a visit-level aggregation
-            step consolidateDiaSourceTable and tract-level aggregation steps
+            These DIA aggregation steps include tract-level aggregation steps
             drpAssociation and drpDiaCalculation.
```
yalsayyad, line 119: Now that the description isn't responsible for explaining which are tract-level and which are visit level, we don't have to call out all the tasks in this step. We should also include a recommendation on the data query: include tract. (It's less important here that people include tract in the data query because diaSources will only exist where a template was found that covers the detector, and "tract" was specified in step 3, but you'll still end up with a ragged edge and bad dia stats.  

"This step includes tract-level aggregation Tasks.  These should be run with explicit 'tract' constraints because otherwise quanta will be created for jobs with only partial visit coverage."

pipelines/imsim/DRP.yaml
```diff
@@ -109,22 +109,25 @@ subsets:
...
-            the data query is recommended. If processing many contiguous tracts
-            running with tract in the dataId will be unavoidable.
-            If run with tract or patch in the data query, drpAssociation and
-            drpDiaCalculation will be complete, but the outputs of
+    step6:
+        subset:
+        - consolidateDiaSourceTable
+        description: >
+            Tasks that can be run together, but only after the 'step1', 'step2',
+            'step3', and 'step4' subsets
```
yalsayyad, line 126: , and 'step5'

pipelines/imsim/DRP.yaml
```diff
@@ -109,22 +109,25 @@ subsets:
...
+
+            This DIA aggregation step includes a visit-level aggregation
```
yalsayyad, line 128: Can also simplify now that all tasks here will be visit level:
 "This step includes visit-level, aggregation tasks. Running without..."

pipelines/imsim/DRP.yaml
hsinfang: Does `consolidateDiaSourceTable` depend on `drpAssociation` and `drpDiaCalculation` (step5)? 
I thought these new `step5` and `step6` aren't blocking either other?

yalsayyad: You're right. They're not blocking each other at the moment. 
I was thinking ahead to running forced phot on the diaObject centroids associated in step5.

PR #356, opened 2021-10-01
DM-30648: Add photodiode calibration class

czwa:
Add script to prepare photodiode data for ingest.


timj:
I think current guidance in the dev guide is to drop the `.py` from script names.

python/lsst/obs/lsst/script/prepare_photodiode.py
```diff
@@ -0,0 +1,111 @@
...
+                        help="Instrument to use for these photodiode data.")
+    return parser
+
+
+def main():
+    args = build_argparser().parse_args()
+
+    # Create output directory or stop.
+    if os.path.exists(args.output_directory):
+        print(f"Output directory {args.output_directory} exists.  Exiting.")
```
mfisherlevine, line 60: I don't think double-spaces are canon in LSST, but I know some people feel strongly.

python/lsst/obs/lsst/script/prepare_photodiode.py
```diff
@@ -0,0 +1,111 @@
...
+        sys.exit(1)
+    else:
+        print(f"Creating output directory {args.output_directory}")
+        os.makedirs(args.output_directory)
+
+    # Determine what files we have to work on.
+    globPath = os.path.join(args.input_directory, args.pattern)
+    inputFiles = glob.glob(globPath)
+    print(globPath, len(inputFiles))
```
mfisherlevine, line 69: Maybe something more like `print(f'Found {len(inputFiles)} in {globPath} for processing`, otherwise this is a little lacking in context.

mfisherlevine: Also probably want an `if len(inputFiles) == 0: exit()` or something before the loop, otherwise I'm not sure what happens when you hit the ingest at the bottom with empty lists. (And it's not that unlikely either, it's easy to mess up your glob.)

python/lsst/obs/lsst/script/prepare_photodiode.py
```diff
@@ -0,0 +1,111 @@
...
+
+    butler = Butler(args.repository)
+
+    ingestFilenames = []
+    ingestInstrument = []
+    ingestExposureId = []
+    for rawPhotodiode in inputFiles:
+        photodiode = PhotodiodeCalib.readSummitPhotodiode(rawPhotodiode)
```
mfisherlevine, line 77: Feel free to ignore, but I dislike variable names for things that they're not. I'd have called this something like photodiodeCalib, or photodiodeData, etc. To me, that should be an instance of a class that represents a diode itself. But I'm probably just being way too picky here, so do ignore if this makes you roll your eyes!

python/lsst/obs/lsst/script/prepare_photodiode.py
```diff
@@ -0,0 +1,111 @@
...
+        day_obs = photodiode.getMetadata()['day_obs']
+        seq_num = photodiode.getMetadata()['seq_num']
+
+        records = [rec for rec in
+                   butler.registry.queryDimensionRecords("exposure", instrument=args.instrument,
+                                                         where=f"exposure.day_obs={day_obs} AND "
+                                                         f"exposure.seq_num={seq_num}")]
```
mfisherlevine, line 84: More for my education, but is there a reason to do this as a list comp rather than just calling `list()` on it? (Obviously there is if you're filtering, or taking parts of the elements, but seems to be identical to `list()`)

python/lsst/obs/lsst/script/prepare_photodiode.py
```diff
@@ -0,0 +1,111 @@
...
+        if len(records) != 1:
+            print(f"Found {len(records)} for day_obs {day_obs} seq_num {seq_num}! Skipping!")
+            continue
```
mfisherlevine, line 87: I don't know enough about the underlying data to know how bad this situation is, but I wonder, if we've got ambiguities, if this should be an exit (with a printout of what the collision is). Might even be a camera team question though.

python/lsst/obs/lsst/script/prepare_photodiode.py
czwa: That is by far the hardest rule for me to follow.

python/lsst/obs/lsst/script/prepare_photodiode.py
czwa: The case I had in mind was for when the raw data hadn't been ingested yet.  Instead of forcing that to be isolated, this skips it, and provides an ingestTable that contains what can be used.

PR #359, opened 2021-10-10
DM-32129: Add a maximum mean offset threshold for SFM astrometry fit success

laurenam:


config/imsim/calibrate.py
```diff
@@ -35,6 +35,8 @@
 config.photoRefObjLoader.load(os.path.join(configDir, "filterMap.py"))
 
 config.astrometry.doMagnitudeOutlierRejection = True
+# Set threshold above which astrometry will be considered a failure (DM-32129)
+config.astrometry.maxMeanDistanceArcsec = 0.08
```
erykoff, line 39: This is a scatter of ~0.4 pixels, right?  That seems bad to me!  But this includes the scatter in the measurement, and is thus band and s/n dependent, etc. I would still personally advocate for something closer to 0.05.  We lose some few more, but these aren't especially good fits anyway, and we don't have any jointcal afterburner to improve things.

erykoff (approved):

PR #360, opened 2021-10-13
DM-32055: Add tests for afw.geom.getAngleBetweenWcs

mfisherlevine:
Tests cannot live in afw as that can't depend on obs.base or obs.lsst
and we need those for a detector model and the wcs creation method.


tests/test_afwWcsUtil.py
```diff
@@ -0,0 +1,76 @@
...
+        for ra, dec, rot1, rot2, flip in itertools.product(ras, decs, rotAngles1, rotAngles2, flips):
+            nomRotation = lsst.geom.Angle(rot2, lsst.geom.degrees)
+            nominalWcs = createInitialSkyWcsFromBoresight(nomPosition, nomRotation, self.detector, flipX=flip)
+
+            testPoint = lsst.geom.SpherePoint(ra, dec, lsst.geom.radians)
+            testRot = lsst.geom.Angle(rot1, lsst.geom.degrees)
+            testWcs = createInitialSkyWcsFromBoresight(testPoint, testRot, self.detector, flipX=flip)
+
+            result = nominalWcs.getRelativeRotationAngleFromWcs(testWcs).asDegrees()
+            test = (rot2-rot1) % 360
```
erykoff, line 62: Shouldn't there be a space around the `-`?

erykoff (approved):

tests/test_afwWcsUtil.py
laurenam: And, in principle, not around `%` 😉 https://developer.lsst.io/python/style.html#binary-operators-should-be-surrounded-by-a-single-space-except-for

mfisherlevine: You don't know how many times I've just rebased, and I've already kicked off Jenkins 😭

PR #361, opened 2021-10-14
DM-32237: Correct doc: step6 depends on step5 now

hsinfang:
as forcedPhotCcdOnDiaObjects is in step5


hsinfang:
Misunderstanding. This update isn't right.

PR #369, opened 2021-11-08
DM-30284: Accommodations towards Gen2/Gen3 parity and determinism in coaddition

laurenam:


python/lsst/obs/lsst/imsim.py
```diff
@@ -44,10 +50,108 @@ class ImsimMapper(LsstCamMapper):
...
+        #
+        policy = None
+        for yamlFile in self.yamlFileList:
+            policyFile = dafPersist.Policy.defaultPolicyFile(self.packageName, yamlFile, "policy")
+            npolicy = dafPersist.Policy(policyFile)
+
+            if policy is None:
+                policy = npolicy
+            else:
+                policy.merge(npolicy)
```
TallJimbo, line 65: Where did this logic come from?  I'm surprised to see it on this ticket, rather than something the base class does for you.

python/lsst/obs/lsst/testHelper.py
```diff
@@ -117,7 +117,7 @@ def _testCoaddId(self, idName):
         coaddId = self.butler.get(idName, dataId={"tract": 9813, "patch": "3,4",
                                                   "filter": self.butler_get_data.filters["raw"]})
         self.assertIsInstance(coaddId, int)
-        maxbits = self.butler.get(f"{idName}_bits")
+        maxbits = 64 - self.butler.get(f"{idName}_bits")
```
TallJimbo, line 120: `64` is unlikely to be the right number, and hard-coding it here is definitely bad.  If it's like ever other case I tried to fix on [DM-17689](https://jira.lsstcorp.org/browse/DM-17689), you probably want to use `lsst.obs.base.ExposureIdInfo.unusedBits`, `lsst.afw.table.IdFactory.computeReservedFromMaxBits`, or ideally one of the highe-rlevel `ExposureIdInfo` methods that delegate to those.  Those will resolve to using `63` instead, because we need to reserve one bit for the sign since these are signed integers, but we don't want to hard-code that number, either.

TallJimbo (approved):

python/lsst/obs/lsst/imsim.py
laurenam: I don't pretend to understand any of this really, but I needed to adapt this mapper to mimic what happens in Gen3.  I tried doing it in the `LsstCamBaseMapper`, but that broke everything for all the other detectors defined in this repo (and I'm not going to pay any Gen2-related attention to those whatsoever).  The only solution I could find was just to copy a lot of what is in `LsstCamBaseMapper` into this one and make the edits I needed.  While I'm pretty sure some of this is redundant, but I spent more time than seemed worth it trying to whittle it down and having it go horribly wrong, that I just figured we could live with this (hopefully VERY transient) duplication.

python/lsst/obs/lsst/testHelper.py
laurenam: Ok, I'm going to need help with this one.  I only added this here to make the comparison consistent after my removal of the "64 - "s in the bit-setting in `lsstCamMapper` such that the tests still pass.  The effective hard-coding of 64 had always been there, and is/was possibly always wrong (in ways I don't pretend to understand), but I was thinking this is all going away, so it would not be worth any effort to "make it right" given its imminent demise.  I tried playing around with `lsst.obs.base.ExposureIdInfo` but was getting lost in clashing mixture of Gen2/Gen3 conventions.  Do you really think this needs fixing (and, if so...help!), or can we just live with it until it all goes away?

TallJimbo: As for whether we can live with it, the question may be whether a 64 here implies that there's a 64 somewhere in Gen3, if you needed to make this 64 in order to get consistency (and a 64 in Gen3 would be bad).  But it sounds like it may be that you (effectively) replaced a 64 somewhere else with a 63 to _get_ consistency with Gen3, and then that broke some test, and then you fixed the wrong thing when trying to fix the test.

But it's also very weird that you're changing something from `x` to `64 - x`.

I would be more inclined to just start deleting any Gen2-only tests that fail if you don't change this; better to have no test than a test that only passes if the behavior is incorrect.

laurenam: Oh, then I'm missing something deeper here...I thought all of this _was_ all Gen2-only (on account of the butler being `lsst.daf.persistence.Butler` in this `testHelper.py`).  The strange change is to mirror the change I made in `lsstCamMapper` (that removed the `64 -` in the value getting compared here).  If I leave the `testHelper` comparison as is, then I get errors in all of the camera-specific tests along the lines:
```=========================================== short test summary info ===========================================
FAILED tests/test_comCam.py::TestLsstCam::testDeepCoaddId - AssertionError: 31 not less than 31 : compare bi...
FAILED tests/test_comCam.py::TestLsstCam::testDcrCoaddId - AssertionError: 31 not less than 31 : compare bit...
FAILED tests/test_imsim.py::TestImsim::testDcrCoaddId - AssertionError: 37 not less than 25 : compare bit le...
FAILED tests/test_imsim.py::TestImsim::testDcrMergedCoaddId - AssertionError: 32 not less than 20 : compare ...
FAILED tests/test_imsim.py::TestImsim::testDeepCoaddId - AssertionError: 37 not less than 25 : compare bit l...
FAILED tests/test_imsim.py::TestImsim::testDeepMergedCoaddId - AssertionError: 32 not less than 20 : compare...
FAILED tests/test_latiss.py::TestLatiss::testDcrCoaddId - AssertionError: 31 not less than 31 : compare bit ...
FAILED tests/test_latiss.py::TestLatiss::testDeepCoaddId - AssertionError: 31 not less than 31 : compare bit...
FAILED tests/test_lsstCam.py::TestLsstCam::testDcrCoaddId - AssertionError: 31 not less than 31 : compare bi...
FAILED tests/test_lsstCam.py::TestLsstCam::testDeepCoaddId - AssertionError: 31 not less than 31 : compare b...
FAILED tests/test_phosim.py::TestPhosim::testDcrCoaddId - AssertionError: 31 not less than 31 : compare bit ...
FAILED tests/test_phosim.py::TestPhosim::testDeepCoaddId - AssertionError: 31 not less than 31 : compare bit...
FAILED tests/test_ts3.py::TestTs3::testDcrCoaddId - AssertionError: 31 not less than 31 : compare bit length...
FAILED tests/test_ts3.py::TestTs3::testDeepCoaddId - AssertionError: 31 not less than 31 : compare bit lengt...
FAILED tests/test_ts8.py::TestTs8::testDcrCoaddId - AssertionError: 31 not less than 31 : compare bit length...
FAILED tests/test_ts8.py::TestTs8::testDeepCoaddId - AssertionError: 31 not less than 31 : compare bit lengt...
FAILED tests/test_ucdcam.py::TestUcdCam::testDcrCoaddId - AssertionError: 31 not less than 31 : compare bit ...
FAILED tests/test_ucdcam.py::TestUcdCam::testDeepCoaddId - AssertionError: 31 not less than 31 : compare bit...
==================== 18 failed, 352 passed, 55 skipped, 596 warnings in 149.48s (0:02:29) =====================
Global pytest run: failed with 1
```

I could just remove the asserts in `testHelper` that are based on this `maxbits` mess?  (I'm now worried that I'm not properly identifying Gen2-specific tests...)

TallJimbo: I am pretty confident everything in this file is Gen2-only, and that's probably true of any tests that actually use it.  Certainly anything involving a `CameraMapper` is Gen2-only.

laurenam: Ok, so am I best off just removing these test asserts rather than adding in the `64 -`?

TallJimbo: I think so.  I'm happy to take a glance at what you end up removing to confirm.

laurenam: Ok, so this was the the original commit: 
https://github.com/lsst/obs_lsst/commit/fd073d95429bb3d53369eb0608645f69c8016686
and this would be the alternative (as a `fixup` that I will `squash` if it's preferred):
https://github.com/lsst/obs_lsst/commit/67a309c5629b63149da611def6fbdf901bbc8e44
Both "work" (i.e. are likely equally bad!)

TallJimbo: Yeah, alternative seems at least fine, and probably a bit better.  Definitely don't want anyone cargo-culting that `64`.

PR #370, opened 2021-11-12
DM-31997: Apply 180 degree offset due to change in defintions

mfisherlevine:


python/lsst/obs/lsst/translators/latiss.py
```diff
@@ -390,6 +394,10 @@ def fix_header(cls, header, instrument, obsid, filename=None):
                 if header[h]:
                     header[h] += offset
 
+        if date > ROTPA_CONVENTION_180_SWITCH1 and date < ROTPA_CONVENTION_180_SWITCH2:
```
timj, line 397: Can we reverse these tests please? For most of the data we take the date will be greater than `SWITCH1` and the test will be true but for only a small fraction of date will the second test be true. If we do the least common case first this will help us for all the remaining data.

python/lsst/obs/lsst/translators/latiss.py
```diff
@@ -390,6 +394,10 @@ def fix_header(cls, header, instrument, obsid, filename=None):
...
+            header['ROTPA'] = header['ROTPA']-180
```
timj, line 398: Usually we leave a space around the operator.

timj (approved):
Looks good. Minor comments.

PR #373, opened 2021-11-23
DM-32141: Get SFM working for LATISS

mfisherlevine:


config/latiss/calibrate.py
```diff
@@ -0,0 +1,67 @@
+import os
+from lsst.utils import getPackageDir
+from lsst.meas.algorithms import LoadIndexedReferenceObjectsTask, MagnitudeLimit
+from lsst.obs.lsst.filters import LATISS_FILTER_DEFINITIONS
+from lsst.meas.astrom import FitAffineWcsTask
+
+# support independent refacts for astrometry and photometry
+# allow configuration at a single point here, with logic following to allow
```
parejkoj, line 8: "with logic following"? What logic?

config/latiss/calibrate.py
```diff
@@ -0,0 +1,67 @@
...
+# easy switching between Gaia, Pan-STARRS and ATLAS refcats.
+ASTROM_REFCAT_NAME = 'gaia_dr2_20200414'
+PHOTO_REFCAT_NAME = 'gaia_dr2_20200414'
```
parejkoj, line 11: The only reason I can think of to have these as named variables here is because we're still holding on to the vestigial `ref_dataset_name` from gen2. I'm not sure that's a good reason to start doing it this way right now.

parejkoj, line 11: Why not use ATLAS for photometry right now? I'd expect it to be better than gaia.

config/latiss/calibrate.py
```diff
@@ -0,0 +1,67 @@
...
+
+# configure the astrometry
+config.connections.astromRefCat = ASTROM_REFCAT_NAME
+config.astromRefObjLoader.retarget(LoadIndexedReferenceObjectsTask)
+config.astromRefObjLoader.ref_dataset_name = ASTROM_REFCAT_NAME
+config.astromRefObjLoader.anyFilterMapsToThis = 'phot_g_mean'
+config.astromRefObjLoader.filterMap = {}  # TODO: remove after DM-33270
```
parejkoj, line 18: No such TODO: that a won't fix.

However, you won't need any astrometry refcat config once DM-27013 merges; might be worth postponing this until then.

config/latiss/calibrate.py
```diff
@@ -0,0 +1,67 @@
...
+
+# support independent refacts for astrometry and photometry
+# allow configuration at a single point here, with logic following to allow
+# easy switching between Gaia, Pan-STARRS and ATLAS refcats.
+ASTROM_REFCAT_NAME = 'gaia_dr2_20200414'
+PHOTO_REFCAT_NAME = 'gaia_dr2_20200414'
+
+# configure the astrometry
+config.connections.astromRefCat = ASTROM_REFCAT_NAME
+config.astromRefObjLoader.retarget(LoadIndexedReferenceObjectsTask)
```
parejkoj, line 15: Irrelevant: only gen2.

config/latiss/calibrate.py
```diff
@@ -0,0 +1,67 @@
...
+config.astromRefObjLoader.ref_dataset_name = ASTROM_REFCAT_NAME
+config.astromRefObjLoader.anyFilterMapsToThis = 'phot_g_mean'
+config.astromRefObjLoader.filterMap = {}  # TODO: remove after DM-33270
+
+# configure the photometry
+config.connections.photoRefCat = PHOTO_REFCAT_NAME
+config.photoRefObjLoader.retarget(LoadIndexedReferenceObjectsTask)
```
parejkoj, line 22: irrelevant: gen2 only

config/latiss/calibrate.py
```diff
@@ -0,0 +1,67 @@
...
+
+# some logic will be required here once we also want to support ATLAS refcat
+# for photometry
+panStarrsFilterMap = {}
+filts = LATISS_FILTER_DEFINITIONS
+for filt in filts._filters:
+    if len(filt.band) == 1:  # skip 'white' etc
+        panStarrsFilterMap[filt.band] = filt.band
+config.photoRefObjLoader.filterMap = panStarrsFilterMap
+config.photoCal.match.referenceSelection.magLimit.fluxField = "i_flux"
```
parejkoj, line 33: You want to always do refcat selection on `i` band, regardless of which filter you're using?

config/latiss/calibrate.py
```diff
@@ -0,0 +1,67 @@
...
+
+# automatically configure for use with Gaia if we're using Gaia
+if 'gaia' in PHOTO_REFCAT_NAME:
```
parejkoj, line 36: I'd suggest not using gaia at all for photometry, and not including any logic for it here. It's just a bad match, and if we have a better gaia catalog in the future (e.g. a preprocessed DR3 that has some pseudo filters) it will look completely different.

config/latiss/calibrate.py
```diff
@@ -0,0 +1,67 @@
...
+    config.photoRefObjLoader.anyFilterMapsToThis = 'phot_g_mean'
+    config.photoRefObjLoader.filterMap = {}  # TODO: remove after DM-33270
+    config.photoCal.match.referenceSelection.magLimit.fluxField = "phot_g_mean_flux"
+
+config.doDeblend = False
+if "ext_shapeHSM_HsmShapeRegauss" in config.measurement.plugins:
+    # if no deblending has been done this is required to not crash
```
parejkoj, line 43: Why might this be in your list of plugins to cause this problem?

config/latiss/calibrate.py
```diff
@@ -0,0 +1,67 @@
...
+    config.measurement.plugins["ext_shapeHSM_HsmShapeRegauss"].deblendNChild = ""
+
+# we often have very few sources so use affine task
+config.astrometry.wcsFitter.retarget(FitAffineWcsTask)
```
parejkoj, line 47: I do wish we'd switched to that by default... But that's not your problem.

config/latiss/calibrate.py
```diff
@@ -0,0 +1,67 @@
...
+
+# should these ever differ, or should we tie them together with MAXOFFSET?
+MAXOFFSET = 3000
+config.astromRefObjLoader.pixelMargin = 1000
```
parejkoj, line 51: You really want to pad by 1000 pixels? That's a lot. At least document why so many.

config/latiss/calibrate.py
```diff
@@ -0,0 +1,67 @@
...
+config.doDeblend = False
+if "ext_shapeHSM_HsmShapeRegauss" in config.measurement.plugins:
+    # if no deblending has been done this is required to not crash
+    config.measurement.plugins["ext_shapeHSM_HsmShapeRegauss"].deblendNChild = ""
+
+# we often have very few sources so use affine task
+config.astrometry.wcsFitter.retarget(FitAffineWcsTask)
+
+# should these ever differ, or should we tie them together with MAXOFFSET?
+MAXOFFSET = 3000
```
parejkoj, line 50: I think this and the pixelMargin below are distinct: this says "you're allowed to be this far off", which could be a good portion of the detector+margin.

config/latiss/calibrate.py
```diff
@@ -0,0 +1,67 @@
...
+config.astromRefObjLoader.pixelMargin = 1000
+config.astrometry.matcher.maxOffsetPix = MAXOFFSET
+
+# apply a magnitude limit and decrease the SNR limit as we're only a 1.2m
```
parejkoj, line 54: All these comments are sentences, so should start with capital letters and have punctuation.

config/latiss/calibrate.py
```diff
@@ -0,0 +1,67 @@
+import os
+from lsst.utils import getPackageDir
+from lsst.meas.algorithms import LoadIndexedReferenceObjectsTask, MagnitudeLimit
+from lsst.obs.lsst.filters import LATISS_FILTER_DEFINITIONS
+from lsst.meas.astrom import FitAffineWcsTask
+
+# support independent refacts for astrometry and photometry
```
parejkoj, line 7: Capital letters and punctuation.

config/latiss/calibrate.py
```diff
@@ -0,0 +1,67 @@
...
+# apply a magnitude limit and decrease the SNR limit as we're only a 1.2m
+# and frequently take short exposures. Also, allow max rotation while the
+# instrumental repeatability isn't good.
+magLimit = MagnitudeLimit()
+magLimit.minimum = 1
+magLimit.maximum = 18
+config.astrometry.referenceSelector.doMagLimit = True
+config.astrometry.referenceSelector.magLimit = magLimit
+config.astrometry.referenceSelector.magLimit.fluxField = "phot_g_mean_flux"
+config.astrometry.matcher.maxRotationDeg = 5.99
```
parejkoj, line 63: 6 degrees only? Seems small if you're aiming for "we could be wildly off".

config/latiss/calibrate.py
```diff
@@ -0,0 +1,67 @@
...
+config.astrometry.sourceSelector['matcher'].minSnr = 10
+
+# write old-style srcMatchFull
```
parejkoj, line 66: I don't think it's an "old style", it's just the full catalog with everything in it.

config/latiss/characterizeImage.py
```diff
@@ -0,0 +1,28 @@
...
+# with logic following to allow easy switching between Gaia and Pan-STARRS
+REFCAT_NAME = 'gaia'
+# REFCAT_NAME = 'ps1'
+
+if REFCAT_NAME == 'gaia':
+    REFCAT = "gaia_dr2_20200414"
+else:
+    REFCAT = "ps1_pv3_3pi_20170110"
+
+config.refObjLoader.retarget(LoadIndexedReferenceObjectsTask)
```
parejkoj, line 13: irrelevant: gen2 only

config/latiss/characterizeImage.py
```diff
@@ -0,0 +1,28 @@
...
+config.refObjLoader.ref_dataset_name = REFCAT
```
parejkoj, line 14: This is gen2 only as well. I don't think you're configuring to use gaia at all in gen3 here.

config/latiss/characterizeImage.py
```diff
@@ -0,0 +1,28 @@
...
+
+# we are at the other end of the regime
+config.doDeblend = False
```
parejkoj, line 17: other end of what?

config/latiss/characterizeImage.py
```diff
@@ -0,0 +1,28 @@
...
+
+# switch to PCA determiner as it is much more robust when there are very few
+# sources. Increase default FWHM due to plate scale and reduce fluxMin due to
+# short expTimes on a 1.2m telescope. nEigenComponents and spatialOrder set to
+# min values so that we always succeed even with one or two stars.
+config.measurePsf.psfDeterminer = 'pca'
+config.measurePsf.psfDeterminer['pca'].spatialOrder = 0
+config.measurePsf.psfDeterminer['pca'].nEigenComponents = 1
+config.measurePsf.starSelector['objectSize'].fluxMin = 2500
+config.installSimplePsf.width = 81  # why is the default 11?!
```
parejkoj, line 27: "?!" - ??

config/latiss/characterizeImage.py
```diff
@@ -0,0 +1,28 @@
...
+config.installSimplePsf.fwhm = 2.355*2
```
parejkoj, line 28: Where's this number from?

config/latiss/processCcd.py
```diff
@@ -26,8 +26,13 @@
 
 ObsConfigDir = os.path.dirname(__file__)
 
-for sub in ("isr", "charImage", "calibrate"):
-    path = os.path.join(ObsConfigDir, sub + ".py")
-    if os.path.exists(path):
-        getattr(config, sub).load(path)
+# this avoids having to either have the wrong filenames, or have stub files
+# which just load the other file, which we should to stop doing in Gen3.
```
parejkoj, line 30: I don't understand this comment. The stuff under this seems unnecessary too.

I don't think you need `processCcd.py` at all, since all of the relevant classes have their own configs that get loaded appropriately, and there's no ProcessCcdTask in gen3, only the pipeline definition.

config/latiss/transformSourceTable.py
```diff
@@ -0,0 +1,5 @@
+import os
+
+# Use the environment variable to prevent hardcoding of paths
+# into quantum graphs.
+config.functorFile = os.path.join('$OBS_LSST_DIR', 'policy', 'latiss', 'Source.yaml')
```
parejkoj, line 5: You don't need this file at all. As of [DM-28862](https://jira.lsstcorp.org/browse/DM-28862), there's a Source.yaml in pipe_tasks and the transform task uses it by default.

policy/latiss/Source.yaml
```diff
@@ -0,0 +1,392 @@
+funcs:
```
parejkoj, line 1: Don't need this file (DM-28862).

Unless you intentionally want something different from the default, in which case you need to document the header of this file.

parejkoj (changes requested):

config/latiss/calibrate.py
mfisherlevine: Wasn't available when the ticket was done, and still hasn't been tested and I don't think it's available everywhere yet. We will definitely update to it once it's been tested, but that'll be another ticket.

config/latiss/calibrate.py
mfisherlevine: Happy to drop a TODO to say fix it once it does, but this has been waiting for too long already and I don't want to block on something else, I've already suffered the drp_pipe rebase here 🙂

config/latiss/calibrate.py
mfisherlevine: I forget why I did that tbh. I will have to ask others to test what happens en masse if we change that, but it was there for a reason.

config/latiss/calibrate.py
mfisherlevine: I know it is, but it's much better than nothing if you're just trying to get the plumbing working, and again we didn't (and still don't quite) have ATLAS. Once that's there, and we confirm it covers the whole of the sky we care about I'll be happy to remove.

config/latiss/calibrate.py
mfisherlevine: Don't know, that was just the default ones, and this is apparently necessary if you're turning deblending off. There might be a ticket to fix that? But I don't know what it is.

config/latiss/calibrate.py
mfisherlevine: Seems very small _indeed._ But it's the max allowed by the config, which is disasterous, but also life.

config/latiss/characterizeImage.py
mfisherlevine: The regime in which one might worry about blending 🙂

config/latiss/characterizeImage.py
mfisherlevine: That's the number of pixels in the PSF model, as far as I can work out. 11 seems absurdly small and causes weird problems. I can't quite understand how this works for anyone.

config/latiss/characterizeImage.py
mfisherlevine: I think/hope it's just doubling a 1" PSF. Anyway, it was just emperical twiddling, and the fact that the platescale is 2x  the LSST nominal.

config/latiss/processCcd.py
mfisherlevine: This ticket was originally for getting it working for Gen2 and Gen3, hence all the Gen2 stuff.

config/latiss/transformSourceTable.py
mfisherlevine: I think it's not the same. IIRC we had to make a bunch of changes to make it work here because we're not running everything.

policy/latiss/Source.yaml
mfisherlevine: As above.

config/latiss/calibrate.py
edennihy: I have confirmed the new ATLAS-REFCAT2 (https://jira.lsstcorp.org/browse/DM-33444) functionally for photometric calibration, but it still requires adding the ticket collection to the input collections (e.g. 'refcats/DM-33444') and I'm still verifying it vs Pan-STARRS calibration. I'm happy to put in the ticket and make this change once that work is done.

policy/latiss/Source.yaml
parejkoj: If you do need a separate file, please document why at the top (if you don't remember the details, at least diff it against the current version in pipe_tasks as a reminder).

policy/latiss/Source.yaml
```diff
@@ -0,0 +1,392 @@
...
+    # apNann: Replaced by raw Aperture instFluxes in flags section below
+    # apMeanSb:  Replaced by raw Aperture instFluxes in flags section below
+    # apMeanSbErr:  Replaced by raw Aperture instFluxes in flags section below
+    extendedness:
+        functor: Column
+        args: base_ClassificationExtendedness_value
+flags:
+#    - base_LocalPhotoCalib
+#    - base_LocalPhotoCalib_flag
+#    - base_LocalPhotoCalibErr
```
parejkoj, line 290: You need these to get calibrated fluxes in the output sourceTables.

policy/latiss/Source.yaml
parejkoj: Reiterating my comment above: If you really do need a separate file here, you need to document what has changed: Erik and I were just trying to figure out why the fluxes were incorrect.

Is there a way to do a "subtraction" from an existing yaml file, so that the difference with the standard version is explicit? Then as you get more things to work, you can just remove those lines from this file.

config/latiss/calibrate.py
edennihy: Confirmed ATLAS-REFCAT2 calibration, updated in 6f7b6a7

policy/latiss/Source.yaml
edennihy: File removed per ce10ce4, I confirmed we could use the default Source.yaml in pipe_tasks after turning on deblend in config.calibrate.py

config/latiss/transformSourceTable.py
edennihy: File removed per ce10ce4, I confirmed we could use the default Source.yaml in pipe_tasks after turning on deblend in config.calibrate.py

policy/latiss/Source.yaml
edennihy: File removed per ce10ce4

config/latiss/calibrate.py
edennihy: updated TO-DO to reference DM-27013 instead

config/latiss/calibrate.py
parejkoj: Document why such a large pixelMargin, please.

config/latiss/processCcd.py
edennihy: Removed in [50b7ad4]

config/latiss/calibrate.py
```diff
@@ -0,0 +1,59 @@
...
+config.connections.photoRefCat = PHOTO_REFCAT_NAME
+config.photoRefObjLoader.retarget(LoadIndexedReferenceObjectsTask)
+config.photoRefObjLoader.ref_dataset_name = PHOTO_REFCAT_NAME
+
+# configure the filter map for atlas refcat2
+atlasFilterMap = {}
+filts = LATISS_FILTER_DEFINITIONS
+for filt in filts._filters:
+    if len(filt.band) == 1:  # skip 'white' etc
+        atlasFilterMap[filt.band] = filt.band
```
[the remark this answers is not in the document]
edennihy: There are additional filters that we exclude with this map are not matched in the ref catalog. E.g. without this map the task fails with 
`lsst.ctrl.mpexec.singleQuantumExecutor ERROR: Execution of task 'calibrate' on quantum {instrument: 'LATISS', detector: 0, visit: 2022050400240, ...} failed. Exception RuntimeError: Unknown reference filter lsst_u_smeared_flux
`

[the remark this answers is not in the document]
timj: what is an "lsst_u_smeared_flux" filter? That's not an actual filter.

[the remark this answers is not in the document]
parejkoj: If the goal is to not photometrically calibrate certain filters, I think that should be done when running the tasks, not abusing the filterMap in this way. I'm surprised that you don't get error when you've specified a filterMap like this (`r=r`, `g=g`, etc.): I'd expect to get a `RuntimeError: Could not find flux field(s) ...` with a filterMap that only contains those things.

The error you're quoting above should only happen if you have `lsst_u_smeared` in the filterMap.

Is the goal to run some kind of minimal photometric calibration on the "non-standard" filters, or to not run any on them?

config/latiss/characterizeImage.py
edennihy: This is actually the default setting for characterizeImage so I will remove.

config/latiss/calibrate.py
edennihy: Retarget removed.

config/latiss/characterizeImage.py
edennihy: This is still needed to satisfy the Gen2/Gen3 consistency check.

edennihy: We have been happily operating with just Gaia here (which is the default), so I don't think this is needed anymore. I will remove this block.

config/latiss/calibrate.py
edennihy: We have to keep the `ref_dataset_name` in until consistency check is removed, so I'll leave this in for now.

parejkoj:
@timj : why did you close this PR without merging it?

PR #376, opened 2021-12-07
DM-32840: Fix bad time to another bad value, but one that translates

mfisherlevine:


corrections/LATISS-AT_O_20200128_000200.yaml
```diff
@@ -0,0 +1,3 @@
+DATE-OBS : '1970-01-01T00:00:01.000'
```
timj, line 1: I assume this means that the DATE and MJD-OBS headers are correct in this file? So they hit the logic where the date starts with 1970 and are replaced. YAML supports a comment field so you could add a quick comment here that says any 1970 date is fine and this triggers the fallback to DATE/MJD-OBS headers.

mfisherlevine: Sure, I think that's correct. To be honest, I just made it match the previous files, because they work, and this one doesn't, and having an obviously-default-like date seemed pretty unobjectionable.
I can add a comment if you like though.

mfisherlevine: Thanks - let me know if you're happy with the comment I've put in and I'll merge if so.

timj (approved):

PR #380, opened 2021-12-29
DM-30891: Replace pipelines with backwards-compatibility stubs or remove them.

TallJimbo:


timj:
This ticket may need some coordination with #386 which also touches the pipelines YAML.

TallJimbo:
Thanks for the heads-up; hopefully @parejkoj 's note that the config change makes the pipeline change unnecessary holds up so there won't be anything to do here, but I'll keep an eye on that ticket.

kfindeisen (approved):

PR #382, opened 2022-01-11
DM-33215: Add support for PROGRAM and REASON to translators

timj:
* REASON applies to all translators.
* PROGRAM applies to LATISS, LSSTCam and ComCam.


edennihy (commented):
These changes were confirmed to work with a local test case on 2021-01-12.

PR #386, opened 2022-01-20
DM-32625: Activate guards on SFM PSF quality for inclusion in coaddition for LSSTCam-imSim

laurenam:


erykoff (approved):
I do have a question why this config is set in the imsim DRP.yaml but the equivalent isn't set in the same place for obs_subaru?

laurenam:
The current defaults are the HSC values, so they don’t need overriding.

pipelines/imsim/DRP.yaml
```diff
@@ -28,6 +28,11 @@ tasks:
             match.refObjLoader.ref_dataset_name: 'cal_ref_cat_2_2'
             python: >
                 config.match.refObjLoader.filterMap = {band: 'lsst_%s_smeared' % (band) for band in 'ugrizy'};
+    makeWarp:
+        class: lsst.pipe.tasks.makeCoaddTempExp.MakeWarpTask
+        config:
+            select.maxEllipResidual: 0.0045
+            select.maxScaledSizeScatter: 0.006
```
parejkoj, line 35: If you've set it in the instrument-specific config files, you don't need to set it in the pipeline yaml.

config/imsim/makeWarp.py
```diff
@@ -0,0 +1,25 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+
+# Set thresholds for PSF fidelity of visit/detector to get included in coadd
+config.select.maxEllipResidual = 0.0045
```
parejkoj, line 24: That comment doesn't say why these specific values were chosen.

pipelines/imsim/DRP.yaml
laurenam: Good point.  I've removed the pipelines overrides and double-checked with 
```
$ pipetask build -p $OBS_LSST_DIR/pipelines/imsim/DRP.yaml#makeWarp --instrument='LSSTCam_imSim' --show config=*select.max*
### Configuration for task `makeWarp'
# Maximum median ellipticity residual
config.select.maxEllipResidual=0.0045

# Maximum scatter in the size residuals
config.select.maxSizeScatter=None

# Maximum scatter in the size residuals, scaled by the median size
config.select.maxScaledSizeScatter=0.006
```

config/imsim/makeWarp.py
laurenam: Details added.

config/imsim/coaddDriver.py
```diff
@@ -0,0 +1,24 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+
```
parejkoj, line 22: Should we have a note that this is gen2-only and can be removed once gen2 is gone?

parejkoj (commented):
Do we really need copyright boilerplate at the top of all of our config files? It makes them much harder to read.

config/imsim/coaddDriver.py
laurenam: Gaarhh, sorry, only saw this after merging.  I think I copy-pasted this from the upper level config (the copyleft is indeed prolly overkill...) Removal of this and all other “driver” config files should happen pretty naturally with gen2 removal, so I think we’re ok without an explicit note.

PR #390, opened 2022-02-14
DM-33078: Define ingest path for photodiode data

czwa:
This ticket adds a new butler cli command, `ingest-photodiode` that reads the plain text files, obtains the matching exposureId, and ingests the `IsrCalib` version.


python/lsst/obs/lsst/ingestPhotodiode.py
```diff
@@ -0,0 +1,209 @@
...
+)
+from lsst.ip.isr import PhotodiodeCalib
+from lsst.obs.base import makeTransferChoiceField
+from lsst.obs.base.formatters.fitsGeneric import FitsGenericFormatter
+from lsst.pex.config import Config
+from lsst.pipe.base import Task
+from lsst.resources import ResourcePath
+
+
+__all__ = ('PhotodiodeIngestConfig', 'PhotodiodeIngestTask')
```
timj, line 41: This needs to be moved up above the other imports.

https://developer.lsst.io/python/style.html?highlight=__all__#standard-code-order-should-be-followed

python/lsst/obs/lsst/__init__.py
```diff
@@ -25,3 +25,4 @@
 from .filters import *
 from ._instrument import *
 from ._fitsHeader import *
+from .ingestPhotodiode import *
```
timj, line 28: Convention suggests that this file should be `_ingestPhotodiode.py` since you are not generally expecting people to do `from lsst.obs.lsst.ingestPhotodiode import X`. Granted we did not do this for `ingest.py` (which should really be called `_ingestRaw.py`) but using the underscore ensures that people will not be doing such a `from` import.

python/lsst/obs/lsst/cli/cmd/commands.py
```diff
@@ -0,0 +1,59 @@
...
+@locations_argument(help="LOCATIONS specifies files to ingest and/or locations to search for files.",
+                    required=True)
+@regex_option(default=r"Photodiode_Readings.*\b",
+              help="Regex string used to find photodiode data in directories listed in LOCATIONS.")
+@config_option(metavar="TEXT=TEXT", multiple=True)
+@config_file_option(type=click.Path(exists=True, writable=False, file_okay=True, dir_okay=False))
+@run_option(required=False)
+@transfer_option()
+@click.option(
+    "--track-file-attrs/--no-track-file-attrs",
```
timj, line 50: I suppose this means I should move this option into reusable code...

python/lsst/obs/lsst/ingestPhotodiode.py
```diff
@@ -0,0 +1,209 @@
...
+            "IsrCalib",
+            universe=self.butler.registry.dimensions,
+        )
+
+    def __init__(self, butler, instrument, config=None, **kwargs):
+        super().__init__(config, **kwargs)
+        self.butler = butler
+        self.universe = self.butler.registry.dimensions
+        self.datasetType = self.getDatasetType()
+        self.progress = Progress("obs.lsst.PhotodiodeIngestTask")
```
timj, line 84: This should use logging convention and so should start with `lsst` (or you could use `self.log.name` which would return something like `lsst.photodiodeIngest`).

python/lsst/obs/lsst/cli/cmd/commands.py
```diff
@@ -0,0 +1,59 @@
...
+from lsst.daf.butler.cli.utils import ButlerCommand
+from ... import script
+
+
+@click.command(cls=ButlerCommand, short_help="Ingest photodiode data.")
+@repo_argument(required=True)
+@instrument_argument(required=True, help="INSTRUMENT is the name of the instrument to use.")
+@locations_argument(help="LOCATIONS specifies files to ingest and/or locations to search for files.",
+                    required=True)
+@regex_option(default=r"Photodiode_Readings.*\b",
```
timj, line 43: This is a different default to that used by the ingest task `run` method. Why?

python/lsst/obs/lsst/cli/cmd/commands.py
```diff
@@ -0,0 +1,59 @@
...
+              help="Regex string used to find photodiode data in directories listed in LOCATIONS.")
```
timj, line 44: Consider having the regex as a variable so that you can include it in the help message.

python/lsst/obs/lsst/ingestPhotodiode.py
```diff
@@ -0,0 +1,209 @@
...
+            URIs to the files to be ingested.
+        pool : `multiprocessing.Pool` optional
+            If not `None`, a process pool with which to
+            parallelize some operations.
+        processes : `int`, optional
+            The number of processes to use.  Ignored if ``pool`` is `None`.
+        run : `str`, optional
+            Name of the RUN-type collection to write to,
+            overriding the default derived from the instrument
+            name
```
timj, line 104: Period to end the sentence.

python/lsst/obs/lsst/ingestPhotodiode.py
```diff
@@ -0,0 +1,209 @@
...
+        else:
+            mode = DatasetIdGenEnum.UNIQUE
+
+        refs = []
+        for inputFile in files:
+            # Convert the file into the right class.
+            if inputFile.isLocal:
+                calib = PhotodiodeCalib.readTwoColumnPhotodiodeData(inputFile.path)
+            else:
+                print("Non-local file.")
```
timj, line 148: The print here should say what file is being skipped and should also have a `continue` after it but in fact this entire if statement should be something like:
```python
with input_file.as_local() as local_file:
    calib = PhotodiodeCalib.readTwoColumnPhotodiodeData(inputFile.path)
```
If you had a constructor that took the content of the file as bytes then you could do `content = input_file.read()` and read all the data (which would even work from S3).

python/lsst/obs/lsst/ingestPhotodiode.py
```diff
@@ -0,0 +1,209 @@
...
+                                                                             where=whereClause,
+                                                                             bind={"dayObs": dayObs,
+                                                                                   "seqNum": seqNum})]
+
+            nRecords = len(exposureRecords)
+            if nRecords == 1:
+                exposureId = exposureRecords[0].id
+                calib.updateMetadata(camera=self.camera, exposure=exposureId)
+            elif nRecords == 0:
+                self.log.warn("Skipping instrument %s and dayObs/seqNum %d %d: no exposures found.",
```
timj, line 166: `self.log.warning`

python/lsst/obs/lsst/ingestPhotodiode.py
```diff
@@ -0,0 +1,209 @@
...
+                              instrumentName, dayObs, seqNum)
+                continue
+            else:
+                raise RuntimeError("Multiple exposure entries found for instrument %s and "
```
timj, line 170: This needs to be an f-string. The %s is not being expanded.

python/lsst/obs/lsst/ingestPhotodiode.py
```diff
@@ -0,0 +1,209 @@
...
+            )
+
+            # If this already exists, we should skip it and continue.
+            existing = {
+                ref.dataId
+                for ref in self.butler.registry.queryDatasets(self.datasetType, collections=[run],
+                                                              dataId=dataId)
+            }
+            if existing:
+                self.log.warn("Skipping instrument %s and dayObs/seqNum %d %d: already exists in run %s.",
```
timj, line 187: `self.log.warning`

timj: Someone may get annoyed if we warn for thousands of files and may ask for this to be a debug but with a warning summarizing all the skips reported at the end.

python/lsst/obs/lsst/ingestPhotodiode.py
```diff
@@ -0,0 +1,209 @@
...
+                              instrumentName, dayObs, seqNum, run)
+                continue
+
+            # Ingest must work from a file, but we can't use the
+            # original, as we've added new metadata and reformatted
+            # it.  Write it to a temp file that we can use to ingest.
+            tempFile = tempfile.mktemp() + ".fits"
```
timj, line 194:
```python
with ResourcePath.temporary_uri(suffix=".fits") as tmpFile:
    calib.writeFits(tmpFile.ospath)
```
or else use `tempfile.NamedTemporaryFile`. This then removes the need to the explicit unlink at the end (which won't happen if there is an error). `FileDataset` can take a `ResourcePath` as `path` argument.

python/lsst/obs/lsst/ingestPhotodiode.py
```diff
@@ -0,0 +1,209 @@
...
+            calib.writeFits(tempFile)
+
+            ref = DatasetRef(self.datasetType, dataId)
+            dataset = FileDataset(path=tempFile, refs=ref, formatter=FitsGenericFormatter)
+
+            # No try, as if this fails, we should stop.
+            self.butler.ingest(dataset, transfer=self.config.transfer, run=run,
```
timj, line 201: Because you are ingesting from a temp file you are not going to be able to support anything other than `copy` for transfer mode. Once we get to a situation where the camera writes out photodiode files in the same format we use (is that ever planned) then in theory things like `link` and `direct` can be allowed.

tests/test_ingest.py
```diff
@@ -127,6 +129,34 @@ class LSSTCamIngestTestCase(IngestTestBase, lsst.utils.tests.TestCase):
...
+        result = runner.invoke(
+            butlerCli,
+            [
+                "ingest-photodiode",
+                self.root,
+                self.instrumentClassName,
+                pdPath,
+            ],
+        )
+        self.assertEqual(result.exit_code, 0, f"output: {result.output} exception: {result.exception}")
```
timj, line 157: You should check that the file can be obtained with `butler.get()` after ingest.

tests/test_ingest.py
```diff
@@ -127,6 +129,34 @@ class LSSTCamIngestTestCase(IngestTestBase, lsst.utils.tests.TestCase):
...
+class LSSTCamPhotodiodeIngestTestCase(IngestTestBase, lsst.utils.tests.TestCase):
+
+    curatedCalibrationDatasetTypes = ("camera",)
+    instrumentClassName = "lsst.obs.lsst.LsstCam"
+    ingestDir = TESTDIR
+    file = os.path.join(DATAROOT, "lsstCam", "raw", "2021-12-12",
+                        "30211212000310", "30211212000310-R22-S22-det098.fits")
+    dataIds = [dict(instrument="LSSTCam", exposure=3021121200310, detector=98)]
+    filterLabel = lsst.afw.image.FilterLabel(physical="SDSSi", band="i")
+    pdPath = os.path.join(DATAROOT, "lsstCam", "raw", "2021-12-12",
```
timj, line 141: Consider not giving a full path and letting the ingest code find the file in just the raw directory.

timj (approved):
Looks great. There are some things I'd like fixed though.

python/lsst/obs/lsst/ingestPhotodiode.py
czwa: I've added a `config.validate` check for this.

python/lsst/obs/lsst/_ingestPhotodiode.py
```diff
@@ -0,0 +1,210 @@
...
+        else:
+            mode = DatasetIdGenEnum.UNIQUE
+
+        refs = []
+        numExisting = 0
+        for inputFile in files:
+            # Convert the file into the right class.
+            with inputFile.as_local() as localFile:
+                calib = PhotodiodeCalib.readTwoColumnPhotodiodeData(localFile.path)
+            calib.writeFits("/tmp/czw.fits")
```
timj, line 146: This seems like a debug that has been left behind.

python/lsst/obs/lsst/_ingestPhotodiode.py
```diff
@@ -0,0 +1,210 @@
...
+            }
+            if existing:
+                self.log.debug("Skipping instrument %s and dayObs/seqNum %d %d: already exists in run %s.",
+                               instrumentName, dayObs, seqNum, run)
+                numExisting += 1
+                continue
+
+            # Ingest must work from a file, but we can't use the
+            # original, as we've added new metadata and reformatted
+            # it.  Write it to a temp file that we can use to ingest.
```
timj, line 192: Now that I think about it, why can't we `butler.put` the calib? It supports writeFits and the FitsGenericFormatter should work so it should be fine.

timj: Is it because eventually you want to do a `direct` ingest? If that's the case then maybe we should mention that in the comment.

czwa: I've updated this comment. I do hope we can have these files be directly ingested in the future.

PR #396, opened 2022-04-22
DM-33942: Support CURINDEX/MAXINDEX/SIMULATE headers for translation

timj:
Used by lsst/astro_metadata_translator#55


python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -834,3 +834,17 @@ def to_group_counter_end(self):
...
+
+    @cache_translation
+    def to_has_simulated_content(self):
+        # Check all the simulation flags.
+        # We do not know all the simulation flags that we may have so
+        # must check every header key. Ideally HIERARCH SIMULATE would
+        # be a hierarchical header so _header["SIMULATE"] would return
+        # everything.
+        for k, v in self._header.items():
+            if "SIMULATE" in k and v:
```
TallJimbo, line 846: Did you really mean for the test to be "if 'SIMULATE' is a substring of k and v is truthy"?  Or was this supposed to be check on whether 'SIMULATE' is a substring of v as well?

I'm also a little bothered that this one metadata lookup is O(N) in the size of the whole header; it's certainly bad news if we have many such things.  But maybe it's all beneath our notice compared to the I/O that comes along with header translation.

TallJimbo (approved):

config/datasetIngest-gen3.py
```diff
@@ -1,7 +1 @@
 # Config override for lsst.ap.verify.Gen3DatasetIngestTask
```
mfisherlevine, line 1: Does this file need to keep existing for some reason? Seems like the single line comment and no code could maybe just go?

python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -802,3 +802,57 @@ def to_boresight_airmass(self):
         log.warning("%s: Unable to determine airmass of a science observation, returning 1.",
                     self._log_prefix)
         return 1.0
+
+    @cache_translation
+    def to_group_counter_start(self):
```
mfisherlevine, line 807: All three of these methods could use a quick docstring.

python/lsst/obs/lsst/translators/lsst.py
mfisherlevine: It's a _long_ story, but I suspect that actually _is_ exactly what he means. There's been a lot of negotiating about cards being/not being present incicating things and what blank values should mean. I think though, given that this really isn't obvious why it's like this, plus the high probability of needing to change this in the future when people change their minds, that a comment explaining this is necessary, so that future people don't "fix" this as a logic bug, or are baffled upon seeing it.

python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -802,3 +802,57 @@ def to_boresight_airmass(self):
...
+        for k, v in self._header.items():
+            if "SIMULATE" in k and v:
+                return True
+
+        # if the controller is H then the data are simulated.
+        ctrlr_key = "CONTRLLR"
+        if self.is_key_ok(ctrlr_key):
+            controller = self._header[ctrlr_key]
+            self._used_these_cards(ctrlr_key)
+            if controller in "HPQ":
```
mfisherlevine, line 854: The comment says "if it's H" but then this test has a bunch of other values. I don't know anything about this, but is this right? Or should the comment be expanded to include these values too?

tests/test_translator.py
```diff
@@ -587,6 +647,72 @@ def test_latiss_translator(self):
...
+                           detector_exposure_id=2022040500348,
+                           detector_group="RXX",
+                           detector_name="S00",
+                           detector_num=0,
+                           detector_serial="ITL-3800C-068",
+                           exposure_id=2022040500348,
+                           exposure_group="2022-04-06T02:58:07.181",
+                           exposure_time=30.0*u.s,
+                           group_counter_end=349,
+                           group_counter_start=348,
```
mfisherlevine, line 664: This and the next one are the only places where group_counter_end != group_counter_start, and they're only separated by 1. Given that (I think) we are using (or will be using) this to group more than 2 exposures, maybe a test for a wider range would be good? Or am I confusing general groupId stuff with the visit system stuff here?

config/datasetIngest-gen3.py
timj: I'm sure it can probably go, it's just that I've run into problems in the past with ap_verify and was a bit too nervous to blow it away. @kfindeisen can you remind me what jenkins job I need to run to make sure I haven't broken something if I remove this file?

python/lsst/obs/lsst/translators/lsst.py
timj: Normally the idea is to read the docstring from MetadataTranslator and we try to have docstring inheritance enabled.

python/lsst/obs/lsst/translators/lsst.py
timj: H controller means simulated phosim. P and Q are other test stand controller modes that imply simulation but I don't think we've ever used them.

python/lsst/obs/lsst/translators/lsst.py
timj: I've updated the comment to be clearer. I don't think there's any way to not have to scan all the headers short of having them sorted and trying to find `HIERARCH SIMULATE* and/or `SIMULATE*` headers in some clever way.

PR #398, opened 2022-05-11
DM-34752: Move non-generic characterize/calibrate configs out of obs_lsst/config

parejkoj:


config/imsim/calibrate.py
```diff
@@ -24,18 +24,34 @@
...
+    refObjLoader.load(os.path.join(configDir, 'filterMap.py'))
+    refObjLoader.ref_dataset_name = 'cal_ref_cat'
+    # Use the filterMap instead of the "any" filter.
+    refObjLoader.anyFilterMapsToThis = None
+
+config.connections.astromRefCat = "cal_ref_cat"
+config.connections.photoRefCat = "cal_ref_cat"
+
+# No color term in simulation at the moment
+config.photoCal.applyColorTerms = False
```
mfisherlevine, line 42: @edennihy I suspect this will need bringing over to the other ticket if you don't have colour terms, and/or want the behaviour to stay the same.

mfisherlevine: Obviously it should then go in LATISS' config, not the base config though.

config/apertures.py
```diff
@@ -25,6 +25,7 @@
 # 'config' should be a SourceMeasurementConfig
 
 config.plugins.names |= ["base_CircularApertureFlux"]
+# TODO: these numbers assume an HSC plate scale!
```
mfisherlevine, line 28: Ticket number for the TODO (sorry, I couldn't help myself!)

config/calibrate.py
```diff
@@ -35,33 +35,9 @@
...
-config.photoCal.applyColorTerms = False
-config.photoCal.match.referenceSelection.doMagLimit = True
-config.photoCal.match.referenceSelection.magLimit.fluxField = "lsst_i_smeared_flux"
-config.photoCal.match.referenceSelection.magLimit.maximum = 22.0
-# select only stars for photometry calibration
-config.photoCal.match.sourceSelection.unresolved.maximum = 0.5
-
-# Demand astrometry and photoCal succeed
-config.requireAstrometry = True
-config.requirePhotoCal = True
```
mfisherlevine, line 63: @edennihy we should see how we had these before, but this is also worth considering.

mfisherlevine (approved):

config/calibrate.py
parejkoj: These are the calibrate defaults.

config/imsim/calibrate.py
parejkoj: I think the default should be fine: it only tries to apply something if the `colorterms` dict has been specified (which defaults to None).

config/calibrate.py
edennihy: We were using the calibrate defaults for `config.photoCal.applyColorTerms = False, config.requireAstrometry = True, config.requirePhotoCal = True` so I'll pull them into the latiss config. We are manually defining the other limits.

config/imsim/calibrate.py
edennihy: I'd like to bring it over just to be more explicit.

config/calibrate.py
parejkoj: What I'm saying is: LATISS doesn't need to set anything for those, because those are the Task defaults. Don't put anything in the LATISS config.

PR #402, opened 2022-05-24
DM-34919: Remove gen2 support

timj:


python/lsst/obs/lsst/testHelper.py
```diff
@@ -62,7 +62,13 @@ def setUpClass(cls):
...
             raise unittest.SkipTest(f"Data directory {cls.data_dir} does not exist.")
 
-        cls._butler = lsst.daf.butler.Butler(cls.data_dir)
+        instrument = cls.getInstrument()
+        # Assume the test repos use the defaults.
+        collections = [instrument.makeUnboundedCalibrationRunName(),
+                       instrument.makeDefaultRawIngestRunName(),
+                       instrument.makeCalibrationCollectionName()]
+        cls._butler = lsst.daf.butler.Butler(cls.data_dir, collections=collections,
+                                             instrument=instrument.getName())
```
parejkoj, line 71: All of the above is instrument agnostic, so it could go into `obs_base:butler_tests.setup_*`, I think?

tests/test_comCam.py
```diff
@@ -28,18 +27,23 @@
...
 
 from lsst.obs.lsst.testHelper import ObsLsstButlerTests, ObsLsstObsBaseOverrides
+from lsst.obs.lsst import LsstComCam
 
 
 class TestLsstCam(ObsLsstObsBaseOverrides, ObsLsstButlerTests):
     instrumentDir = "comCam"
 
+    @classmethod
+    def getInstrument(cls):
```
parejkoj, line 37: We should probably make this an abstract method on the parent class, since it's needed there.

parejkoj (commented):
How are ccdExposureIds handled in gen3? Those and detector names are the removals here, so I hope we're testing the modern interface for that somewhere?

tests/test_comCam.py
timj: I think it's only needed in the obs_lsst testHelper. I tried adding something to obs_base but the multiple inheritance (3 different classes) got confused.

timj: Ah. I misread what you said. Yes, testHelper could have it indeed.

timj:
ccdExposureId is not anything to do with butler so the test is no longer relevant.

detector names are all handled transparently by the detector dimension record in butler (which comes from the ObservationInfo) so there is no specialist parsing involved, unlike gen2. The tests asking for a dataset by raft and name_in_raft show that it works.

python/lsst/obs/lsst/testHelper.py
timj: Only if the obs_base tests are dropping gen2 and we think any of these tests are worth keeping.

PR #408, opened 2022-07-07
Add FOCUSZ into LSST translators

jbkalmbach:


timj (approved):
This looks fine. I see we have some latiss test headers that triggers the code path.

PR #410, opened 2022-07-08
DM-35524: Change default TS8 filter from NONE to unknown

timj:
This matches a "valid" TS8 filter definition.


jchiang87 (approved):
LGTM.  I tried this on the TS8 data at issue after setting
```
TSTAND = 'TS8'
INSTRUME = 'LSST-TS8'
```
in the PHDU, and the ingest ran fine.

PR #411, opened 2022-07-11
DM-35528: Update TS8 info in obs_lsst for FITS header changes and CCD replacement in slot S00 RTM-004

jchiang87:


python/lsst/obs/lsst/translators/ts8.py
```diff
@@ -242,6 +242,6 @@ def to_exposure_id(self):
     @cache_translation
     def to_observation_id(self):
         # Docstring will be inherited. Property defined in properties.py
-        filename = self._header["FILENAME"]
-        self._used_these_cards("FILENAME")
-        return filename[:filename.rfind(".")]
+        observation_id = self._header["OBSID"]
```
timj, line 245: This will break with old TS8 data (there is a reason I didn't use OBSID before) and I don't think it's going to pass the tests. I think you will need to do:

```python
if self._is_key_ok("OBSID"):
    observation_id = self._header["OBSID"]
    self._used_these_cards("OBSID")
    return observation_id
filename = self._header["FILENAME"]
...
```
and leave the old FILENAME code in place.

You likely should add a new TS8 header to the tests/header directory and add a test for it.

timj (changes requested):
I think the obsid code needs to be tweaked so it handles old data. Please add a new test header as well so we can be sure it works.

timj (approved):

PR #416, opened 2022-08-05
DM-8889: Add detector z-offsets

jmeyers314:


RobertLuptonTheGood (commented):
Looks good.

At the risk of attracting complaints about the general lack of docs on yamlCamera, where is it documented that offsets may be specified with 2- or 3- values (with [2] defaulting to 0)?

jmeyers314:
> where is it documented that offsets may be specified with 2- or 3- values (with [2] defaulting to 0)?

Added https://github.com/lsst/obs_lsst/pull/416/files#diff-a66f10959b31a508e4f363ab528d9c3e55f99de92d1c5ce8f14eabe266404d42R37-R39.

PR #419, opened 2022-08-15
DM-35902: Remove getAmpImage

timj:
Gen2-specific.


RobertLuptonTheGood (commented):
Looks good, given that gen3 addresses this natively.

However, I am worried that there may be client code (e.g. camera team notebooks) that use this API.  Should we provide a helper function, or at least a function that says, "Please use `butler.get(..., amp=1)`" (or whatever the syntax is)?

timj:
If any existing camera notebooks are using this then those notebooks will be breaking long before they get to this part of the code. They can no longer access a gen2 butler so they can't be at the point where they have anything to pass to this function. I don't think adding a message in this function will ever trigger for existing notebooks.

PR #422, opened 2022-09-14
Add new filters that will go in the new 65mm filter wheel for LATISS.

patrickingraham:


patrickingraham:
Passed CI as well..
https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/37292/pipeline

timj (approved):

PR #423, opened 2022-09-14
DM-36169: merge extra hard-coded named headers in rawFormatter

weatherhead99:
In current LATISS data (and likely future comcam & LSSTCAM data), the FITS files may contain extra HDUs "REB_COND" and "CONFIG_COND", which describe some sensor state that does not make sense to be included per-amplifier. This change checks if those HDUs are present, and if they are includes them in the metadata.

### DIFF
--- a/python/lsst/obs/lsst/rawFormatter.py
+++ b/python/lsst/obs/lsst/rawFormatter.py
@@ -55,6 +55,9 @@ class LsstCamRawFormatter(FitsRawFormatterBase):
     filterDefinitions = LsstCam.filterDefinitions
     _instrument = LsstCam
 
+    # These named HDUs' headers will be checked for and added to metadata.
+    _extraFitsHeaders = ["REB_COND", "CONFIG_COND"]
+
     def readMetadata(self):
         """Read all header metadata directly into a PropertyList.
 
@@ -68,15 +71,29 @@ class LsstCamRawFormatter(FitsRawFormatterBase):
         """
         file = self.fileDescriptor.location.path
         phdu = lsst.afw.fits.readMetadata(file, 0)
+
         if "INHERIT" in phdu:
             # Trust the inheritance flag
-            return super().readMetadata()
+            base_md = super().readMetadata()
 
         # Merge ourselves
-        md = merge_headers([phdu, lsst.afw.fits.readMetadata(file)],
-                           mode="overwrite")
-        fix_header(md)
-        return md
+        else:
+            base_md = merge_headers([phdu, lsst.afw.fits.readMetadata(file)],
+                                    mode="overwrite")
+
+        ehdrs = []
+        for hduname in self._extraFitsHeaders:
+            try:
+                ehdr = lsst.afw.fits.readMetadata(file, hduname)
+            except lsst.afw.fits.FitsError:
+                # We can ignore this, the header doesn't exist in this file.
+                continue
+            else:
+                ehdrs.append(ehdr)
+
+        final_md = merge_headers([base_md] + ehdrs, mode="first")
+        fix_header(final_md)
+        return final_md
 
     def getDetector(self, id):
         in_detector = self._instrument.getCamera()[id]


python/lsst/obs/lsst/rawFormatter.py
```diff
@@ -68,15 +70,29 @@ def readMetadata(self):
...
+        else:
+            base_md = merge_headers([phdu, lsst.afw.fits.readMetadata(file)],
+                                    mode="overwrite")
+            fix_header(base_md)
+
+        ehdrs = []
+        for hduname in self._extraFitsHeaders:
+            try:
+                ehdr = lsst.afw.fits.readMetadata(file, hduname)
+                ehdrs.append(ehdr)
```
timj, line 88: Ordinarily we try to minimize the content of the try block to only the code that we are wanting to protect. This append can go outside the try. You can use an `else` block for it if you like.

python/lsst/obs/lsst/rawFormatter.py
```diff
@@ -68,15 +70,29 @@ def readMetadata(self):
...
+            except lsst.afw.fits.FitsError:
+                # can ignore, the header just doesn't exist in this file
+                continue
+
+        final_md = merge_headers([base_md] + ehdrs)
```
timj, line 93: I assume none of the headers in these extensions clash with primary headers? There are multiple options in `merge_headers` for specifying how to handle repeat headers.

python/lsst/obs/lsst/rawFormatter.py
```diff
@@ -68,15 +70,29 @@ def readMetadata(self):
...
+                                    mode="overwrite")
+            fix_header(base_md)
+
+        ehdrs = []
+        for hduname in self._extraFitsHeaders:
+            try:
+                ehdr = lsst.afw.fits.readMetadata(file, hduname)
+                ehdrs.append(ehdr)
+            except lsst.afw.fits.FitsError:
+                # can ignore, the header just doesn't exist in this file
```
timj, line 90: (minor comment: Python style (ie Guido) wants full sentences as comments so "# Can" and finishing with a period).

python/lsst/obs/lsst/rawFormatter.py
```diff
@@ -55,6 +55,8 @@ class LsstCamRawFormatter(FitsRawFormatterBase):
     filterDefinitions = LsstCam.filterDefinitions
     _instrument = LsstCam
 
+    _extraFitsHeaders = ["REB_COND", "CONFIG_COND"]
```
timj, line 58: Can you add a comment here explaining what this is for?

timj (approved):
Thanks for doing this. Some minor comments.

I did notice when looking around the code base that we have a `python/lsst/obs/lsst/_fitsHeader.py` that purports to read the header. I think this was a helper function for gen2 at some point. I probably should make the formatter use that code at some point.

python/lsst/obs/lsst/rawFormatter.py
```diff
@@ -68,15 +70,29 @@ def readMetadata(self):
...
 
         # Merge ourselves
-        md = merge_headers([phdu, lsst.afw.fits.readMetadata(file)],
-                           mode="overwrite")
-        fix_header(md)
-        return md
+        else:
+            base_md = merge_headers([phdu, lsst.afw.fits.readMetadata(file)],
+                                    mode="overwrite")
+            fix_header(base_md)
```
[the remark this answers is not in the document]
weatherhead99: understood on the `fix_header` call. 

On the subject of the amplifier headers, I think there is some stuff potentially of interest in there. Possibly not for this ticket. 

However, the implementation of doing that would probably just look like iterating through all the HDUs and pulling stuff out, so would be different to this one.

python/lsst/obs/lsst/rawFormatter.py
weatherhead99: I doubt it but will check with some real data and look into it.

PR #428, opened 2022-11-11
DM-34989: Update obs_lsst to fix common problems.

czwa:
* Set size of ITL overscan region to 64 pixels.
* Remove DETSEC log warning that will always trigger due to external value mismatches.
* Set the overscan.fitType default to MEDIAN_PER_ROW.

### DIFF
--- a/python/lsst/obs/lsst/assembly.py
+++ b/python/lsst/obs/lsst/assembly.py
@@ -168,8 +168,14 @@ def fixAmpGeometry(inAmp, bbox, metadata, logCmd=None):
     datasec = bboxFromIraf(d["DATASEC"]) if "DATASEC" in d else None
     biassec = bboxFromIraf(d["BIASSEC"]) if "BIASSEC" in d else None
 
-    if detsec and outAmp.getBBox() != detsec:
-        logCmd("DETSEC doesn't match (%s != %s)", outAmp.getBBox(), detsec)
+    # 2022-11-11: There is a known issue that the header DETSEC have
+    # the y-axis values flipped between the C0x and C1x entries.  This
+    # is incorrect, and disagrees with the cameraGeom values.
+    # DM-36115 contains additional details.  This test has been
+    # disabled to remove useless warnings until that is resolved.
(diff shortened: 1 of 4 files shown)


python/lsst/obs/lsst/assembly.py
```diff
@@ -168,8 +168,11 @@ def logCmd(*args):
     datasec = bboxFromIraf(d["DATASEC"]) if "DATASEC" in d else None
     biassec = bboxFromIraf(d["BIASSEC"]) if "BIASSEC" in d else None
 
-    if detsec and outAmp.getBBox() != detsec:
-        logCmd("DETSEC doesn't match (%s != %s)", outAmp.getBBox(), detsec)
+    # 2022-11-11: There is a known issue that the header DETSEC does
+    # not match the cameraGeom value: DM-36115.  Disabling this test
+    # to remove useless warnings until that is resolved.
+    # if detsec and outAmp.getBBox() != detsec:
+    #     logCmd("DETSEC doesn't match (%s != %s)", outAmp.getBBox(), detsec)
```
erykoff, line 175: I do not understand this comment.  Or rather, is this still necessary with the update to the `cameraHeader.yaml`?  Or is this about past images?

czwa: This is all images.  The `DETSEC` recorded in the headers swaps the y-position of the upper and lower sets of amplifiers.  Craig has the exact values on the referenced ticket:
https://jira.lsstcorp.org/browse/DM-36115?focusedCommentId=856790&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-856790

I don't think it's expected to be fixed in the near future, so I removed the test to keep the log noise down to prevent confusion.

erykoff: Okay, can you rephrase the comment so that it's clear that the header DETSEC is incorrect?  This comment and the linked issue title are ambiguous about whether the problem is the header or cameraGeom.

erykoff (approved):
See my suggested clarification of the code comment.

PR #436, opened 2023-02-07
DM-37890: Add blue300lpmm_quadnotch1 filter

patrickingraham:


timj (approved):
I see you changed your mind on the name of the filter. Make sure the final name is what is in the header. Also, please squash the two commits together before merging.

patrickingraham:
I had to shorten the name because tests failed... somewhere there is some length limit imposed.

timj:
physical filter has a butler limit of 32 characters. I think the problem here is that LATISS filters are the combination of grating and filter because the butler data model doesn't understand grating as a distinct concept.

PR #439, opened 2023-02-20
DM-38005: Update astrometry reference matcher configs for LATISS

laurenam:
Now that the pointing accruacy and stability has improved for AutTel, the bbox padding for reference catalog loading can be greatly reduced. This has the benefit of avoiding many very bad fits where the matcher locked onto sources way outside of the true field for the exposure.

### DIFF
--- a/config/latiss/calibrate.py
+++ b/config/latiss/calibrate.py
@@ -7,22 +7,29 @@ config.photoCal.match.referenceSelection.magLimit.fluxField = "r_flux"
 # We often have very few sources due to smaller aperture so use affine task. 
 config.astrometry.wcsFitter.retarget(FitAffineWcsTask)
 
-# Increase maxoffset as AUXTEL pointing can be unreliable. Between Feb2020-Mar2022 we saw offsets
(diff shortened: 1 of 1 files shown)


config/latiss/calibrate.py
```diff
@@ -11,18 +11,21 @@
 # of up to 4 arcmin, which translates to 2400pix. We choose 3000 as a conservative limit. pixelMargin 
 # also increased to ensure refcat overlap when we see large pointing offsets considering a 3000pix 
 # offset and detector half-width of 2100pix.
```
edennihy, line 13: Could you update these comments to reflect the new choices? It may also be worth adding a warning that these values were validated using data from December 2022, and that data taken prior to May 2022 may require more conservative limits.

edennihy (approved):

config/latiss/calibrate.py
laurenam: Comment added.  Let me know if it captures things adequately!

edennihy: That looks good, thanks!

PR #452, opened 2023-05-22
DM-39327: added config files for analysis_tools associatedSourcesTractAnalysis

psferguson:
This adds the two config files needed to run photometric repeatability focal plane plots.


jeffcarlin (approved):
Looks good, and I confirmed that it runs fine for me. I don't see AstrometricRepeatability plots showing up in the Plot Navigator, but the PhotometricRepeatability ones are there.

PR #455, opened 2023-05-25
DM-39415: Fix TS8 exposure ID calculation

timj:
We have already ingested TS8 data into a butler using the old timestamp exposure_id calculation. It is very disruptive to change this because it will require the entire repository be remade from scratch. Furthermore, the fix to DATE-OBS broke all the historical calculations where the exposure_id was being derived from the end of the integration by mistake.

This PR retains LSSTCam exposure_id for > 2023-05-24 and uses the original DATE-OBS value to calculate the exposure ID for older data.

Corrects the changes introduced in #453


python/lsst/obs/lsst/translators/ts8.py
```diff
@@ -24,6 +24,9 @@
 
 log = logging.getLogger(__name__)
 
+_EXPOSURE_ID_DATE_CHANGE = Time("2023-05-24T12:00:00", format="isot", scale="utc")
```
timj, line 27: @ktlim do we have an exact time for when we started ingesting with modern LSSTCam exposure ID records that I need to still be calculating the same way? Maybe I can work it out directly from /repo/ir2.

timj: And the answer is the first observation on 2023-05-25 (2023-05-25 17:30:34.75401).

jchiang87 (approved):
Looks good to me.  However, I think there is a separate issue regarding how these changes would affect the behavior of `butler register-instrument --update`.

timj:
You mean the visit_max stuff? The `--update` option doesn't fix that?

timj:
I think I can fix the max_exposure_id problem pretty quickly.

timj:
Okay. I think I've fixed the max exposure ID.

jchiang87 (approved):
Looks good (nice use of regexps).

python/lsst/obs/lsst/translators/ts8.py
```diff
@@ -153,6 +161,22 @@ def compute_detector_exposure_id(cls, exposure_id, detector_num):
...
+        The TS8 implementation is non-standard because TS8 data can create
+        two different forms of exposure_id based on the date but we need
+        the largest form to be the one returned.
+
+        Returns
+        -------
+        max_exposure_id : `int`
+            The maximum value.
+        """
+        max_date = "2050-12-31T23:59.999"
```
ktlim, line 177: Shouldn't this have 59 seconds too?

timj: Yes, Ha!  Except the original calculation actually had that and it is in fact used in the TS8 calculation like that if we want to reproduce the original value. The other instruments work because they only take the YYYY-MM-DD part.

PR #458, opened 2023-06-16
DM-39585: Test ingest of ECSV format photodiode data

czwa:


python/lsst/obs/lsst/_ingestPhotodiode.py
```diff
@@ -143,34 +143,67 @@ def run(self, locations, run=None, file_filter=r".*Photodiode_Readings.*txt",
...
+                    # can't use defaults.
+                    instrumentName = self.instrument.getName()
+
+                obsId = calib.getMetadata()['obsId']
+                whereClause = "exposure.obs_id=obsId"
+                exposureRecords = [rec for rec in registry.queryDimensionRecords("exposure",
+                                                                                 instrument=instrumentName,
+                                                                                 where=whereClause,
+                                                                                 bind={"obsId": obsId})]
+                idTuple = (obsId)
```
timj, line 173:
```suggestion
                idTuple = (obsId,)
```
otherwise it's not a tuple and it looks like you want a tuple.

```python
>>> x = ("id")
>>> y = ("id",)
>>> type(x)
<class 'str'>
>>> type(y)
<class 'tuple'>
```
Given you only log the contents you probably don't need a tuple but then consider a different variable name.

python/lsst/obs/lsst/_ingestPhotodiode.py
```diff
@@ -143,34 +143,67 @@ def run(self, locations, run=None, file_filter=r".*Photodiode_Readings.*txt",
...
+            if calibType == "full":
+                instrumentName = calib.getMetadata().get('INSTRUME')
+                if instrumentName is None:
+                    # The field is populated by the calib class, so we
+                    # can't use defaults.
+                    instrumentName = self.instrument.getName()
+
+                obsId = calib.getMetadata()['obsId']
+                whereClause = "exposure.obs_id=obsId"
+                exposureRecords = [rec for rec in registry.queryDimensionRecords("exposure",
```
timj, line 169: I'd rather this query code wasn't duplicated. You can set up the `instrumentName`, `bind` and `whereClause` variables in the `if` branch and then do the query outside.

python/lsst/obs/lsst/_ingestPhotodiode.py
```diff
@@ -143,34 +143,67 @@ def run(self, locations, run=None, file_filter=r".*Photodiode_Readings.*txt",
...
+                calibType = "full"
+            except Exception:
+                # Try reading as a two-column file
+                with inputFile.as_local() as localFile:
+                    calib = PhotodiodeCalib.readTwoColumnPhotodiodeData(localFile.ospath)
+                calibType = "two-column"
+
+            # Get exposure records
+            if calibType == "full":
+                instrumentName = calib.getMetadata().get('INSTRUME')
```
timj, line 161: We have "new" ECSV files where this is not set? Or is this something that Tony is adding in the future? What happens is the required `instrument` parameter for `__init__` does not match the value from the header?

python/lsst/obs/lsst/_ingestPhotodiode.py
```diff
@@ -206,8 +239,8 @@ def run(self, locations, run=None, file_filter=r".*Photodiode_Readings.*txt",
                 # No try, as if this fails, we should stop.
                 self.butler.ingest(dataset, transfer=self.config.transfer,
                                    record_validation_info=track_file_attrs)
-                self.log.info("Photodiode %s:%d (%d/%d) ingested successfully", instrumentName, exposureId,
-                              dayObs, seqNum)
+                self.log.info("Photodiode %s:%d (%s) ingested successfully", instrumentName, exposureId,
+                              idTuple)
```
timj, line 243: In theory since the exposure record is available you have the choice of being consistent here and always using day_obs/seq_num or always using obs_id (also for the log message above).

python/lsst/obs/lsst/cli/cmd/commands.py
```diff
@@ -35,7 +35,7 @@
 from ... import script
 
 
-defaultRegex = r".*Photodiode_Readings.*txt\b"
+defaultRegex = r".*Photodiode_Readings.*txt|.*photodiode.ecsv\b"
```
timj, line 38: I wonder if `$` could be used instead of `\b`? Something like

```suggestion
defaultRegex = r"(Photodiode_Readings_[0-9_]+\.txt|photodiode\.ecsv)$"
```
? The leading `.*` aren't really needed).

python/lsst/obs/lsst/_ingestPhotodiode.py
```diff
@@ -88,7 +88,7 @@ def __init__(self, butler, instrument, config=None, **kwargs):
         self.instrument = instrument
         self.camera = self.instrument.getCamera()
 
-    def run(self, locations, run=None, file_filter=r".*Photodiode_Readings.*txt",
+    def run(self, locations, run=None, file_filter=r".*Photodiode_Readings.*txt|.*_photodiode.ecsv",
```
timj, line 91: See discussion on regex at the end.

timj (approved):
Some minor cleanups requested but I will approve.

python/lsst/obs/lsst/_ingestPhotodiode.py
czwa: Changed to simply be `logId` (of `str` or `tuple` as needed), as this is just for the log messages.

python/lsst/obs/lsst/_ingestPhotodiode.py
czwa: I've requested that this be changed.  Once that's done, then as part of the DM-39694 I plan to remove the `instrument` argument (removing it from the `__init__`), so photodiode ingest will determine the instrument solely from the input files.

PR #462, opened 2023-07-12
DM-39999: Add UC Davis Beam simulator ITL camera

Snyder005:
Add UC Davis Beam Simulator ITL Camera to obs_lsst.


python/lsst/obs/lsst/translators/lsst_ucdcam.py
```diff
@@ -96,191 +69,14 @@ def can_translate(cls, header, filename=None):
             Header to convert to standardized form.
         filename : `str`, optional
             Name of file being translated.
-
```
timj, line 99: This line needs to go back I think.

tests/test_translator.py
```diff
@@ -1233,60 +1233,6 @@ def test_ts8_translator(self):
                     warnings.simplefilter("ignore", VerifyWarning)
                     self.assertObservationInfoFromYaml(filename, dir=self.datadir, **expected)
 
-    def test_ucdcam_translator(self):
```
timj, line 1236: There needs to be a check of the translator. Please add a couple of test headers back (with different chips) and ensure that you are getting the expected values out.

timj (changes requested):
The Jira ticket does not make it clear but this PR is going to result in the older data not being importable into gen3 butlers. Is everyone okay with that? Today the "SAO" test stand data can be ingested but next week it won't be. As an archivist this seems wrong to me but if people are deleting all that data then that's going to be okay.

We can't merge this PR without some test headers being added to ensure that these files can be translated properly.

tests/test_translator.py
Snyder005: I added some test headers for the E2V and ITL CCDs. As I understand it I know need a translator test object where I define the `test_data` dictionary which then is compared to the header information I put in the test header YAML files?

timj: The code deleted from this file should show you what form the replacement code needs to be. You can get the expected output by running `astrometadata -p lsst.obs.lsst.translators translate` on the test yaml file.

I take it there is agreement that none of the 2018 code should ever be read into butler again?

Snyder005: I believe the translator has been written completely now.  I was able to run scons without any failures.

Snyder005:
Added test headers and wrote the test translator function for the new LSST-UCDCam camera definition.

timj:
Can you confirm that the old data that this code worked with before will never needed to be looked at again? I'm yet to get a formal statement that those data are now irrelevant and nothing will need to look at them.

Snyder005:
Yes, the old data was only used by a few of us here at Davis and was never made available at USDF. We don't intend to use it within the DM Butler framework again.

python/lsst/obs/lsst/_instrument.py
```diff
@@ -379,3 +346,35 @@ def getRawFormatter(self, dataId):
         # local import to prevent circular dependency
         from .rawFormatter import LatissRawFormatter
         return LatissRawFormatter
+
+
+class LsstUCDCam(LsstCam):
```
timj, line 351: Can you please move this class back to where it was (above LsstTS3) so it is more obvious what has changed and what has stayed the same? I think the only thing that has changed is the filter definition (other than you changing the order of the properties).

python/lsst/obs/lsst/translators/lsst_ucdcam.py
```diff
@@ -103,184 +76,10 @@ def can_translate(cls, header, filename=None):
...
-        run : `str`
-            YYYY-MM-DD string corresponding to the date of observation.
-        """
-        # Get a copy so that we can edit the default formatting
-        date = self.to_datetime_begin().copy()
-        date.format = "iso"
-        date.out_subfmt = "date"  # YYYY-MM-DD format
-        return str(date)
+        if "INSTRUME" in header:
+            instrume = header["INSTRUME"].lower()
```
timj, line 80:
```suggestion
            if header["INSTRUME"].lower() in (cls.supported_instrument.lower(), "LSST-UCDCam-ITL".lower()):
```
?

python/lsst/obs/lsst/translators/lsst_ucdcam.py
```diff
@@ -103,184 +76,10 @@ def can_translate(cls, header, filename=None):
...
+            if instrume == cls.supported_instrument.lower():
+                return True
+            elif instrume == "LSST-UCDCam-ITL".lower():
```
timj, line 83: If ever we learn that this is a hotspot we could replace the string with  `"lsst-ucdcam-itl"`...

tests/test_gen3.py
```diff
@@ -157,10 +157,6 @@ def testTs3(self):
         self.checkInstrumentWithRegistry(LsstTS3,
                                          "ts3/raw/2016-07-22/201607220607067-R071-S00-det071.fits")
 
-    def testUcdCam(self):
-        self.checkInstrumentWithRegistry(LsstUCDCam,
-                                         "ucd/raw/2018-12-05/20181205233148-S00-det000.fits")
```
timj, line 162: You haven't got a stripped FITS file you want to test you can ingest into a butler?

timj (approved):
Thanks for the changes. I've made some minor comments but I would like the class moved back where it was if possible to show that the changes are small.

tests/test_gen3.py
Snyder005: By "stripped FITS" file, how exactly is this modified form a regular image file.  It sounds like some data is removed, to decrease the file size.

timj: Yes. See eg https://pipelines.lsst.io/v/weekly/modules/lsst.obs.lsst/testing.html#butler-tests for example script to do that.

PR #464, opened 2023-07-24
DM-38986: Update gain, read noise, saturation values for LSSTCam

jchiang87:


python/lsst/obs/lsst/script/updateFromEOTest.py
```diff
@@ -132,15 +131,6 @@ def updateFromEOTest(RXXDir, raftNames, updatedData):
         with open(RXXFile) as fd:
             raftNameData = yaml.load(fd, yaml.CSafeLoader)
 
-        shutil.copyfile(RXXFile, RXXFile + "~")  # make a backup, although we also have git
```
plazas, line 135: It's not necessary to do this back up anymore?

plazas (approved):
Looks good! Thanks! 
Just a clarification: In Jira you commented that dead amplifiers were given average values for gain and noise. Why is this the case? If they're dead, shouldn't we flag/mask them and not use them?

python/lsst/obs/lsst/script/updateFromEOTest.py
jchiang87: The unaltered files are already kept track of by git.  And if this code is run more than once, as I had to do while understanding that the workaround for corner rafts wasn't needed, the existing backup files get overwritten with altered data, so they're really not useful at that point.

PR #465, opened 2023-08-14
DM-39858: Add imSim-specific overrides for CalibrateImageTask

parejkoj:


config/imsim/isr.py
```diff
@@ -22,5 +22,4 @@
 """
 imsim-specific overrides for IsrTask
 """
-config.doCrosstalk = True
```
kfindeisen, line 25: If the `obs_lsst` default changed to `False`, would you want this one to follow suit or remain at `True`?

kfindeisen (approved):
Looks good; minor question about defaults.

config/imsim/isr.py
parejkoj: Good question: I could see it going either way, but we know we need to do crosstalk correction in all the LSSTCam chips (simulated and real), so I don't foresee it ever being set to False in the lowest-level instrument override.

PR #466, opened 2023-08-15
DM-40301: Update configurations to use large 35 pixel aperture for LATISS calibration.

erykoff:
The default 12 pixel aperture is much too small for the LATISS platescale.

### DIFF
--- a/config/latiss/calibrate.py
+++ b/config/latiss/calibrate.py
@@ -42,3 +42,7 @@ config.astrometry.doMagnitudeOutlierRejection = True
 config.doWriteMatchesDenormalized = True
 
 config.measurement.plugins["base_Jacobian"].pixelScale = 0.1
+
+# Set the default aperture as appropriate for the LATISS plate scale.
+config.measurement.slots.apFlux='base_CircularApertureFlux_35_0'
+config.measurement.slots.calibFlux='base_CircularApertureFlux_35_0'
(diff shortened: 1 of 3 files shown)


config/latiss/calibrate.py
```diff
@@ -42,3 +42,7 @@
 config.doWriteMatchesDenormalized = True
 
 config.measurement.plugins["base_Jacobian"].pixelScale = 0.1
+
+# Set the default aperture as appropriate for the LATISS platescale.
```
laurenam, line 46: Huh...I always think of "plate scale" as two words (here and elsewhere in this PR if you decided to change it, but I'm not insisting!)

config/latiss/isolatedStarAssociation.py
```diff
@@ -0,0 +1,27 @@
+# Override calibration apertures as appropriate for LATISS platescale.
+config.inst_flux_field='apFlux_35_0_instFlux'
+
+config.extra_columns = [
+    'x',
+    'y',
+    'apFlux_50_0_instFlux',
+    'apFlux_50_0_instFluxErr',
+    'apFlux_50_0_flag',
```
laurenam, line 9: Why 50 here?

laurenam (approved):

config/latiss/isolatedStarAssociation.py
erykoff: For testing background offsets, fgcm uses the calibration aperture plus 1 larger.  50 is that one larger.  (And the calibration flux already gets transferred, this is just the "extra" bit.)

PR #467, opened 2023-08-18
DM-40426: Add new LATISS filters collimator and cyl_lens, and new gratings holo4_001 and pinhole mask options. 

edennihy:
Jenkins run -> https://rubin-ci.slac.stanford.edu/job/stack-os-matrix/36/

Note that the PR is organized in 3 separate commits, one for the new filters, one for the new grating, and the third is for the options of the pinhole masks. The pinhole masks (there are two) will only be installed in the grating wheel, but they have configurable options as far as the aperture size and aperture locations. The names were chosen to express all available configurations as it is unknown ahead of time which aperture size/location configuration will be used.

### DIFF
--- a/python/lsst/obs/lsst/filters.py
+++ b/python/lsst/obs/lsst/filters.py
@@ -298,7 +298,8 @@ _latiss_filters = (
 )
 
 # Form a new set of filter definitions from all the explicit gratings
-_latiss_gratings = ("ronchi90lpmm", "ronchi170lpmm", "empty", "unknown", "holo4_003", "blue300lpmm_qn1")
+_latiss_gratings = ("ronchi90lpmm", "ronchi170lpmm", "empty", "unknown",
+                    "holo4_003", "blue300lpmm_qn1", "holo4_001")
 
 # Include the filters without the grating in case someone wants
 # to retrieve a filter by an actual filter name
--- a/python/lsst/obs/lsst/filters.py
+++ b/python/lsst/obs/lsst/filters.py
@@ -298,8 +298,34 @@ _latiss_filters = (
 )
 
 # Form a new set of filter definitions from all the explicit gratings
-_latiss_gratings = ("ronchi90lpmm", "ronchi170lpmm", "empty", "unknown",
-                    "holo4_003", "blue300lpmm_qn1", "holo4_001")
+_latiss_gratings = ("ronchi90lpmm",
+                    "ronchi170lpmm",
+                    "empty",
+                    "unknown",
+                    "holo4_003",
+                    "blue300lpmm_qn1",
+                    "holo4_001",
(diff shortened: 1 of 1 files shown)


python/lsst/obs/lsst/filters.py
```diff
@@ -291,10 +291,41 @@ def addFilter(filter_dict, band, physical_filter):
                      band="z"),
     FilterDefinition(physical_filter="SDSSy_65mm",
                      band="y"),
+    FilterDefinition(physical_filter="collimator",
+                     band="collimator"),
```
timj, line 295: We try not to name bands after filters if there is another option. We have a current list of: diffuser, u, g, r, i, z, y, unknown, notch, white.  Do collimator and cyl_lens have more general band names?

python/lsst/obs/lsst/filters.py
```diff
@@ -291,10 +291,41 @@ def addFilter(filter_dict, band, physical_filter):
...
+                    "pinhole_2_2_0200",
+                    "pinhole_2_2_0100",
+                    "pinhole_2_3_1000",
+                    "pinhole_2_3_0500",
+                    "pinhole_2_3_0200",
+                    "pinhole_2_3_0100",
+                    "pinhole_2_4_1000",
+                    "pinhole_2_4_0500",
+                    "pinhole_2_4_0200",
+                    "pinhole_2_4_0100",
```
timj, line 327: Are the gratings turning up in files with these names or is there a requirement to derive the names from multiple FITS headers in the metadata translator?

timj (commented):
I have a couple of comments before I do the review.

python/lsst/obs/lsst/filters.py
edennihy: They do not have more general band names, but among that list I think either "white" or "unknown" would be the most appropriate as these two elements do not have any intentional attenuation by wavelength.

timj: Maybe use `white` then.

python/lsst/obs/lsst/filters.py
edennihy: The gratings should be turning up in files with one of these names at a time. 

To expand a bit, the plan is to install the pinhole mask into the grating stage instrument with one of these names depending on the pinhole configuration we setup outside of the instrument, before the mask is installed. We chose this list of names to cover all possible configurations as we expect it to change over time. However, that change is always done as part of a grating swap.

python/lsst/obs/lsst/filters.py
edennihy: Recommendation to use `white` adopted and pushed.

timj (approved):
Looks good.

PR #471, opened 2023-09-14
DM-40555: Add configurations for running fgcmcal on LATISS photometric data

erykoff:


hlin730 (approved):
Looks good, see comments in https://jira.lsstcorp.org/browse/DM-40555

PR #472, opened 2023-09-26
DM-40918: Use unified CCS/OCS sequence number for exposure_id

timj:


python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -412,13 +412,20 @@ def compute_exposure_id(dayobs, seqnum, controller=None):
         if seqnum >= 10**_SEQNUM_MAXDIGITS:
             raise ValueError(f"Sequence number ({seqnum}) exceeds limit")
 
+        dayobs = int(dayobs)
+        if dayobs > 20231004 and controller == "C":
```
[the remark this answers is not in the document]
ktlim: Might be nice to move the cast up above the `seqnum` code to where the rest of the `dayobs` handling is.  But the `if` pertains to `controller` so should probably stay here.

Yes, the OODS also uses this ingest code.  I understand that it can be any weekly (although I think OODS and auto-ingest can use dailies if need be) after the Camera has actually switched, but the fewer images that are taken in between the better.  (I suspect that most of the controller-C images taken in between will actually be tests of the system.)

tests/headers/lsstCam-MC_C_20190319_000001_R10_S02.yaml
```diff
@@ -32,7 +32,7 @@ TELESCOP: LSST
 OBSID: MC_C_20190319_000001
 TELCODE: MC
 CONTRLLR: C
-DAYOBS: '20190319'
+DAYOBS: '20290319'
```
ktlim, line 35: Should the file be renamed, or is this also supposed to be a test of using headers over filename components?

tests/test_translator.py
```diff
@@ -45,13 +45,13 @@ def test_lsstCam_translator(self):
                            instrument="LSSTCam",
                            boresight_rotation_coord="unknown",
                            dark_time=0.0*u.s,
-                           detector_exposure_id=0x269200011d,
```
ktlim, line 48: Interesting that this was broken anyway due to 3019 being beyond the 2050 range of the packer.

ktlim (approved):

tests/headers/lsstCam-MC_C_20190319_000001_R10_S02.yaml
timj: If I rename the file I probably should change all the other 2019s in there to 2029. It's partly why the commit was called a "fudge". It's probably better if I copy and rename and add as entirely different entry specifically for testing the exposure IDs otherwise I remove the test and add one later when we get a real unified CCS observation.

python/lsst/obs/lsst/translators/lsst.py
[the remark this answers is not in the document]
timj: @tony-johnson is 20231004 a reasonable date for me to pick if I merge by tomorrow (so would be in this weekly)?

tests/headers/lsstCam-MC_C_20190319_000001_R10_S02.yaml
ktlim: How are you going to tell that a CCS observation was unified?  I thought you can only tell by not seeing a corresponding OCS observation.

ktlim: Or do you mean just taking a real observation after the cutover date?

timj: Yes, I meant taking a real CCS observation after the cutoff and then noting that I got the exposure ID correct for it (no 3_000 offset). I'll leave the existing file as is and copy it and fudge it, making clear it's not a real header.

PR #473, opened 2023-10-05
DM-41025: Add SDSS_65mm filter options to grating slot for LATISS.

edennihy:


ktlim:
Might be nice to filter out combinations that will never be used; right now the nested loops will make every combination.

ktlim:
Also, you have aliases for the filter in the filter wheel (new line 357); do you need aliases for the filter in the grating wheel?

edennihy:
Do you have a suggestion for a quick way to filter out all of the (anticipated) unused combinations?

edennihy:
To your second point, I noticed the SDSS65mm filter definitions above did not have alias definitions, so I wasn't sure if that was needed. Same thing for the afw_name, unless those are being defined elsewhere.

edennihy:
Okay, I added some simple logic to only allow the (empty~SDSS_65mm) combination and a check on the grating alias if needed.

timj:
How many combinations do you end up with after this change?

```
$ butler create tmp
$ butler register-instrument tmp lsst.obs.lsst.Latiss
$ butler query-dimension-records tmp physical_filter
```
?

edennihy:
The total number of filters prior to change is 701, total number after is 707. The majority are due to combinations with the pinhole masks. We could filter some of those as well if we want (for example, order blocking filters will not be used with the pinhole masks).

python/lsst/obs/lsst/filters.py
```diff
@@ -325,6 +325,18 @@ def addFilter(filter_dict, band, physical_filter):
                     "pinhole_2_4_0500",
                     "pinhole_2_4_0200",
                     "pinhole_2_4_0100",
+                    FilterDefinition(physical_filter="SDSSu_65mm",
```
timj, line 328: Can you add a comment above this block explaining why these filters are nominally in the gratings list?

python/lsst/obs/lsst/filters.py
```diff
@@ -325,6 +325,18 @@ def addFilter(filter_dict, band, physical_filter):
...
+                                     band="u"),
+                    FilterDefinition(physical_filter="SDSSg_65mm",
```
timj, line 330: I worry a bit that this duplication of class constructors with the filter list is going to leave us susceptible to typos in the future. Maybe we define a `sdss_filters` list and then use `*sdss_filters` in here and the filter list above? That would guarantee that we really are duplicating.

python/lsst/obs/lsst/filters.py
```diff
@@ -338,17 +350,36 @@ def addFilter(filter_dict, band, physical_filter):
...
             continue
 
+        # The SDSS_65mm filters in the grating position should never be used
+        # with any filter other than "empty", so we skip those as well.
+        if hasattr(grating, "physical_filter") and filter.physical_filter != "empty":
+            continue
+
         # FilterDefinition is a frozen dataclass
-        new_name = FILTER_DELIMITER.join([filter.physical_filter, grating])
+        if hasattr(grating, "physical_filter"):
```
timj, line 359: I think you can do something like:

```python
grating_name = getattr(grating, "physical_filter", grating)
new_name = FILTER_DELIMITER.join([filter.physical_filter, grating_name])
```

python/lsst/obs/lsst/filters.py
```diff
@@ -338,17 +350,36 @@ def addFilter(filter_dict, band, physical_filter):
...
-        combo = FilterDefinition(physical_filter=new_name,
-                                 band=filter.band,
-                                 afw_name=filter.afw_name,
-                                 alias=new_aliases)
+        if hasattr(grating, "physical_filter"):
+            new_aliases = {FILTER_DELIMITER.join([filter.physical_filter, a]) for a in grating.alias}
+        else:
+            new_aliases = {FILTER_DELIMITER.join([a, grating]) for a in filter.alias}
+
+        # For gratings set the band to the band of the filter. If an SDSS_65mm
```
timj, line 370: This commit is far more specific than the code that is written here. The comment should be talking about what happens if the grating is a filter definition. SDSS_65mm is never checked.

python/lsst/obs/lsst/filters.py
```diff
@@ -338,17 +350,36 @@ def addFilter(filter_dict, band, physical_filter):
...
+        # filter is installed in the grating slot, define the band as the band
+        # of the SDSS_65mm filter.
+        if hasattr(grating, "band"):
+            combo = FilterDefinition(physical_filter=new_name,
```
timj, line 374: Since the above code is mainly working out the parameters for these calls, maybe use one `isinstance` instead to calculate the parameters instead of doing lots of hasattr tests?

timj (changes requested):
I think this needs a bit of tidying up.

It's great that you managed to filter out all the combinations. I agree that another pass in the future to remove impossible combinations would be great.

timj (approved):

PR #474, opened 2023-10-23
DM-41267: Enable calculation of weather data in translator

timj:


python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -934,3 +934,34 @@ def to_has_simulated_content(self):
...
+        return None
+
+    @cache_translation
+    def to_pressure(self):
+        key = "PRESSURE"
+        if self.is_key_ok(key):
+            value = self._header[key]
+            # There has been an inconsistency in units for the pressure reading
+            # so we need to adjust for this.
+            if value > 10_000:
```
timj, line 954: If we think this was a one-off blip then it might be more efficient to do this in header fixup. Doing it here makes it fixes it for LSSTCam/ComCam and LATISS but if it's a thing that is in the past and won't change in the future doing the fix just for LATISS and for a restricted date range makes more sense.

mfisherlevine (approved):

python/lsst/obs/lsst/translators/lsst.py
mfisherlevine: I don't have strong feelings either way. I guess maybe it depends what you mean by "more efficient" though. If it's just this `if` statement, then I don't think it even helps, because you still have to check the date anyway, plus it's negligible either way. But therefore perhaps you meant something else.

I personally like this approach, because it should fix it everywhere forever, and feels safe and non-costly, and we never thought this would happen, so it's hard to say it won't happen again, I think. But like I said, I don't feel strongly if you want to make it time-based and remove this general check, but I think it's fine to leave as is, myself.

timj: My point is that putting it here means we check it for the rest of the survey in both LSSTCam and LATISS data. Putting it in the LATISS fix_header limits the check to LATISS data and if we are clever with the date checking we might be able to limit it to commissioning data without needing an extra check.

If the CSCs can randomly change units on us and we expect the units to switch multiple times over the survey we have much bigger problems. cc/ @tribeiro

tribeiro: Certainly, the CSCs will not change units randomly. However, we can't rule out that sensor changes and software bugs may cause data to be published with wrong units. We'll do our best to minimize those, of course.

PR #476, opened 2023-10-25
DM-41235: Update test to use SqlRegistry instead of Registry

andy-slac:


timj (approved):

andy-slac:
@timj, could you make rebase-checker action required for this package?

PR #480, opened 2023-10-31
DM-40955: LATISS psf stamp size is incorrect for the large PSFs

laurenam:


config/latiss/characterizeImage.py
```diff
@@ -15,6 +15,9 @@
 
 # Reduce psfex spatialOrder to 1, this helps ensure success with low numbers of psf candidates.
 config.measurePsf.psfDeterminer["psfex"].spatialOrder = 1
+# Set the default kernel and stamp sizes for PSF modelling appropriate for LATISS
```
erykoff, line 18: US spelling is `modeling` (one `l`).

config/latiss/finalize_characterization.py
```diff
@@ -1,2 +1,8 @@
-# Reduce piff spatialOrder to 1, this helps ensure success with low numbers of psf candidates.
-config.psf_determiner['piff'].spatialOrder = 1
+# Change the PSF determiner to psfex from piff default as the latter will not
+# scale well to the large kernel/stamp sizes for PSF modelling for LATISS data.
```
erykoff, line 2: Also here.

config/latiss/finalize_characterization.py
```diff
@@ -1,2 +1,8 @@
...
+config.psf_determiner.name = "psfex"
+# Reduce spatialOrder to 1, this helps ensure success with low numbers of psf candidates.
+config.psf_determiner["psfex"].spatialOrder = 1
+# Set the default kernel and stamp sizes for PSF modelling appropriate for LATISS.
```
erykoff, line 6: And here.

erykoff (approved):
Minor spelling comments.

laurenam:
I made the update to move the `import` from `finalizeCharacterization` to the `config` here (and added `setupOptional(meas_extensions_psfex`) to `ups/obs_lsst.table`).  Let me know if you want to take another look while Jenkins is running.

ups/obs_lsst.table
```diff
@@ -9,6 +9,7 @@ setupRequired(astro_metadata_translator)
 setupRequired(obs_lsst_data)
 setupRequired(daf_butler)
 setupOptional(meas_extensions_gaap)
+setupOptional(meas_extensions_psfex)
```
erykoff, line 12: I'm never sure what should be optional or required here, and I'm sure this has to be picked up by pipe_tasks currently, but I think it's okay to have this here.

erykoff:
Assuming you confirmed that moving to the config works, then 👍

ups/obs_lsst.table
laurenam: I'm not totally clear on this either!  My reasoning was:
a) we should not rely on it being picked up by another required package (as that could change from underneath us).
b) it's "optional" because it is only actually "required" if processing LATISS data (but I'm not 100% sure that's correct reasoning since it seems that it's crossing a line).

I'll leave it as is for now (but will keep it in mind if we see problems down the road that could be related).

PR #484, opened 2023-11-15
DM-41319: Add configurations for LATISS empty~ filters which may be installed in the second (grating) wheel

erykoff:


config/latiss/colorterms.py
```diff
@@ -24,14 +24,29 @@
             c1=0.0610018330811135,
             c2=-0.0722575356707918,
         ),
+        # empty~i is the same as i~empty.
```
czwa, line 27: Is there a reason to not define the `g`, `r`, and `z` flipped filters now as well?

czwa (approved):
Same comment can be mapped to all three files.  Worst case it just prompts a future ticket, so I'm happy with this as is.

config/latiss/colorterms.py
erykoff: I'm taking a gamble that we won't need to do this, but we can always revisit.  (We'd have to add it to the transmission filters as well as the fgcm lookup table, etc.)

PR #487, opened 2023-11-18
DM-41675:  Update BOT filters and add to TS8 filter set

jchiang87:


tony-johnson (approved):
I checked the set of real filters in the fcs and they look correct. I don't completely understand how we are combining filters to get names like ef_43~nm750 but they look consistent at least.

PR #489, opened 2023-11-24
DM-41838: Update maxScaledSizeScatter thresholds for LSSTCam-imSim

laurenam:


config/imsim/makeWarp.py
```diff
@@ -30,5 +30,9 @@
 # After switching to Piff for PSF estimation, these proved too tight for at
 # least ci_imsim.  The updates just added a bit of padding top of what was
 # needed to get all images in ci_imsim through.
+#
+# The maxScaledSizeScatter has been increased further to reflect the new
+# metric definition (which results in higher values for this metric).  See
+# DM-40668 & DM-41838 for details.
 config.select.maxEllipResidual = 0.005
-config.select.maxScaledSizeScatter = 0.008
+config.select.maxScaledSizeScatter = 0.011
```
parejkoj, line 38: Is this specific to imSim? It seems like something that should be set either in the package defaults (if it's for all cases of makeWarp) or in the DRP ingredients, so it applies to all instruments.

erykoff: These config values depend on typical performance of a given instrument in typical conditions.  The appropriate value for HSC is not the same for DC2 is not the same for LATISS.

erykoff (approved):

PR #494, opened 2023-12-12
DM-42115: Add z, y to fgcmcal LATISS config and fix minimum number of detectors for "good" visits

erykoff:


config/latiss/fgcmFitCycle.py
```diff
@@ -23,32 +23,45 @@
...
 config.sedterms = fgcmcal.SedtermDict()
 config.sedterms.data = {
     "g": fgcmcal.Sedterm(primaryTerm="gr", secondaryTerm="ri", constant=1.5),
     "r": fgcmcal.Sedterm(primaryTerm="gr", secondaryTerm="ri", constant=0.9),
     "i": fgcmcal.Sedterm(primaryTerm="ri", secondaryTerm="gr", constant=0.5,
                          extrapolated=True, primaryBand="i", secondaryBand="r", tertiaryBand="g"),
+    "z": fgcmcal.Sedterm(primaryTerm="iz", secondaryTerm="zy", constant=1.0),
+    "y": fgcmcal.Sedterm(primaryTerm="zy", secondaryTerm="iz", constant=0.25,
+                         extrapolated=True, primaryBand="y", secondaryBand="z",
+                         tertiaryBand="i"),
```
laurenam, line 64: Why do some get `prim/second/ter/iaryBand` entries (just curious)?  Is it worth mentioning something about how the `constant` values are determined?

laurenam (approved):

config/latiss/fgcmFitCycle.py
erykoff: The formulae are in the appendix of the FGCM Burke et al paper, and the constants ... well, I should put my scripts somewhere, but they come from eyeballing synthetic loci actually.  Things are not super sensitive to these constants (at least in any metrics that we looked at in DES)

PR #498, opened 2024-02-12
DM-42287:  Simulated ComCam instrument model for ops-rehearsal-3

jchiang87:


python/lsst/obs/lsst/translators/comCamSim.py
```diff
@@ -0,0 +1,100 @@
...
+        Notes
+        -----
+        Content will be added as needed.
+
+        Corrections are reported as debug level log messages.
+
+        See `~astro_metadata_translator.fix_header` for details of the general
+        process.
+        """
+        modified = False
```
timj, line 82: Could fix that TELESCOP problem described above.

tests/test_packer.py
```diff
@@ -246,6 +248,22 @@ def test_comCam(self):
...
+        # $ butler query-dimension-records data/input/comCamSim exposure \
+        #      --where "instrument='LSSTComCamSim'" --limit 1
+        self.check_rubin_dimension_packer(
+            instrument,
+            is_exposure=False,
+            exposure_id=5024032100720,
+            day_obs=20240321,
+            seq_num=720,
+            detector=4,
+            controller="P"
```
timj, line 264: I am concerned that we are using P controller to mean something other than what we have historically thought of as PLAYBACK. @ktlim  and I discussed this and thought that it should be controller "S" for simulated. "P" is meant to mean rerun real observations exactly as they were taking on the mountain but maybe I'm misunderstanding.

timj (approved):
This looks good modulo my lack of understanding what "P" controller really means these days.

PR #499, opened 2024-02-12
DM-42154: Add config overrides for LATISS multiband to turn off refcat matching.

erykoff:
On this PR I also added in the correct override of the filter map to allow matches to be turned on if we need them (though I don't think they're actually used anywhere).


fred3m (approved):
LGTM

PR #500, opened 2024-02-13
DM-42837: Add LsstCam and ComCam distortion coefficients

jmeyers314:


jchiang87:
Josh:  Can you also add the coefficients to comCamSim?   That code was just merged this afternoon.  Thanks!

jmeyers314:
Do we also want to update `phosim` and `imsim` (which I think should be the same as `LsstCam`)?  Will there be an `LsstCamSim` at some point?

timj:
There is already a `LsstCamSim` but it's called `LsstCam-imSim` (so yes, comcam is inconsistent with that).

jmeyers314:
Ah, okay.  I suspect `LsstCam-imSim` is what I meant by `imsim`.  I.e., [this](https://github.com/lsst/obs_lsst/blob/main/policy/imsim/cameraHeader.yaml#L29).  That's probably what we used for DC2 I guess?  I'm suspicious though, does this still work for current full-camera imsim sims? @jchiang87 ?

timj:
I think there is a whole debate somewhere about "new" imsim simulations. I think we wanted them to be an entirely new camera in obs_lsst and not be LSSTCam or LSSTCam-imSim.

jchiang87:
> does this still work for current full-camera imsim sims? 
I'm not sure what you mean.  I would have guessed that the DC2 era distortions might be different than the ones we'd infer now with the current code, using batoid, etc.

Regarding Tim's point about an entirely new camera model for simulated LSSTCam:   I would like to make one that's different from `LSSTCam-imSim` and I guess I'd call it now simply `LSSTCamSim`.   As far as I can tell, the only reason to keep `LSSTCam-imSim` around as-is is because it's still being used in CI packages and elsewhere (e.g., DP0).

timj:
@jchiang87 I'm fine with a new `LSSTCamSim`. The `LSSTCam-imSim` has to stay around for DP0 compatibility and the headers used in those are wildly different from what a modern simulated LSSTCam would have.

jmeyers314:
> I would have guessed that the DC2 era distortions might be different than the ones we'd infer now with the current code, using batoid, etc.

For DC2, I believe we used a WCS originally derived from phosim, which presumably did raytracing through the same optics model (the as-designed Rubin model) as we're now tracing through with batoid.  It's possible phosim used a different model, or was buggy (though I don't _know_ of any bugs in this part of phosim).  I suppose I should go check a DC2 raw to see if adding the new distortion model to `LSSTCam-imSim` makes an improvement...

jmeyers314:
Okay, here's before and after adding the same distortion coefficients to LSSTCam-imSim as being added to LsstCam.  Comparison is dead-reckoned star positions given the camera model vs the actual image:

Old

old_DC2

New

new_DC2

So not perfect, but a definite improvement.

jmeyers314:
Forgot to mention: above is a DC2 image against the cal_ref_cat_2_2 catalog.

PR #501, opened 2024-02-14
DM-42862: initial DRP pipeine for  ComCamSim

yalsayyad:
- transformObjectCatalog and the
- measurement plugins in MeasureMergedCoaddSourcesTask which apply to all cameras.


erykoff (approved):
I’m an idiot. It’ll work with the final characterization

PR #503, opened 2024-02-21
DM-42971: Move refCat overrides from drp_pipe comcamsim pipelines to here

yalsayyad:


config/comCamSim/analysisToolsPhotometricCatalogMatch.py
```diff
@@ -0,0 +1,4 @@
+"""comCamSim-specific overrides for AstrometricCatalogMatchTask"""
```
hsinfang, line 1: `PhotometricCatalogMatchTask`

hsinfang (approved):

PR #509, opened 2024-03-04
DM-43033: Add fgcmcal configurations for ComCamSim ops rehearsal data

erykoff:


config/comCamSim/fgcmFitCycle.py
```diff
@@ -0,0 +1,108 @@
...
+config.ccdGraySubCcdDict = {"g": True,
+                            "r": True,
+                            "i": True}
+config.ccdGrayFocalPlaneDict = {"g": True,
+                                "r": True,
+                                "i": True}
+config.ccdGrayFocalPlaneFitMinCcd = 3
+config.ccdGrayFocalPlaneChebyshevOrder = 2
+# Do not model the magnitude errors (use errors as reported).
+config.modelMagErrors = True
```
cmsaunders, line 83: Shouldn't this be False, based on the comment above?

cmsaunders (approved):
I tried to check whether the config parameters seem reasonable, but I'm mostly trusting you on the values. I did notice one that seems inconsistent with the associated comment. 
I'd also prefer squashing the commits.

config/comCamSim/fgcmFitCycle.py
erykoff: Or the comment is copied from a test config where we don't have much coverage.  Will fix comment.

PR #514, opened 2024-03-15
DM-43175: Fix translation problem with negative exposure time and no DATE-END

timj:
Without a `DATE-END` the end time is calculated from the begin time and the exposure time. In some cases we deliberately calculate a negative exposure time and this was leading to the end time being before the begin time. Now catch this and set end time to the begin time.


mfisherlevine (approved):
LGTM.

PR #515, opened 2024-03-18
DM-43368: Add TEMP_SET to keywords to check for ISR.

czwa:
Add TEMP_SET to keywords to check.

### DIFF
--- a/config/latiss/isr.py
+++ b/config/latiss/isr.py
@@ -26,5 +26,5 @@ config.doCrosstalk=True
 config.doDefect=True
 config.overscan.fitType="MEDIAN_PER_ROW"
 config.overscan.doParallelOverscan=True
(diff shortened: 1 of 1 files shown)


erykoff (approved):
These are all automatically forwarded into the calib headers, right?  This is just comparing TEMP_SET which is already in the calib header and warning if it doesn't match?

czwa:
They will be automatically forwarded into the combined calibration header only if they are identical in all input exposures.  I don't think this was a decision made with this feature in mind, but rather to ensure there's no ambiguity in the header values that are propagated from the inputs.  For this case, either the inputs all agree (and the keyword appears in the calibration header), or they disagree, the keyword is excluded from the combined calibration, and using that calibration for any processing will result in a warning that the calibration does not have the keyword for comparison (and will raise in the future when we make that switch).

timj:
The header merging scheme can be controlled but at the moment there is no ability to say that headers that differ should be dropped apart from one specific header which should have all values duplicated.

PR #516, opened 2024-03-20
DM-43347: Configure visit-level refcat match analysis tools  tasks for ComCamSim 

yalsayyad:
The  RefCatSourcePhotometricAnalysisTask one, as of now is necessary, but not sufficient.

### DIFF
--- a/config/comCamSim/refCatSourcePhotometricAnalysisTask.py
+++ b/config/comCamSim/refCatSourcePhotometricAnalysisTask.py
@@ -1,0 +1,3 @@
+"""comCamSim-specific overrides for RefCatSourcePhotometricAnalysisTask"""
+
+config.connections.data = "sourceTable_visit_uw_stars_20240228_photoMatch"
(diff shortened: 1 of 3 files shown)


config/comCamSim/refCatSourcePhotometricAnalysisTask.py
```diff
@@ -0,0 +1,3 @@
+"""comCamSim-specific overrides for RefCatSourcePhotometricAnalysisTask"""
+
+config.connections.data = "sourceTable_visit_uw_stars_20240228_photoMatch"
```
ctslater, line 3: I don't know the details here but just to double check, is this supposed to end in `_photoMatch` or `_match`? I don't see this datset type in `/repo/ops-rehearsal-3-prep`.

yalsayyad: This is the one that doesn't run.  After Sophie's ticket in Dec, the astrometry match tasks use the old names and the photometry one adds the word photo or Photometric in front of everything. This is the one I commented on the PR that it doesn't run. Necessary but not sufficient.  I was hoping to get some help from @sr525 during pair coding today. Maybe someone else can get it going in the next week

PR #517, opened 2024-03-23
DM-43436: Enable OBJECT and AZEL translations for LSSTCam

timj:
ComCamSim needs special code because the AZEL values are missing and airmass is inconsistent with the observation date.


timj:
It has been suggested on Slack that I also add header fixup code for comcamsim to copy RA/DEC into [RA|DEC][START|END} since they must be identical in simulated data like this.

This doesn't solve the AZEL problem but does mean that downstream of ISR the header will be filled in.

python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -953,10 +971,8 @@ def to_has_simulated_content(self):
                 return True
 
         # If the controller is H, P, or Q then the data are simulated.
-        ctrlr_key = "CONTRLLR"
-        if self.is_key_ok(ctrlr_key):
-            controller = self._header[ctrlr_key]
-            self._used_these_cards(ctrlr_key)
+        controller = self._get_controller_code()
+        if controller:
             if controller in "HPQ":
```
jchiang87, line 976: Should this test also include controller "S"?

jchiang87 (changes requested):
Looks good, though I think the controller code test for simulated data should include "S".

python/lsst/obs/lsst/translators/lsst.py
timj: @jchiang87  I fixed this but I also added some header fixups for things like RASTART/RAEND and fixed an age-old ComCam translator bug from #167. It's quite a lot of code so I've asked for a re-review.

timj:
I will fix the problem with the day_obs offset calculation on another ticket. It's not important for ops rehearsal since we are not yet using butler universion 6 for that.

jchiang87 (approved):
lgtm.

PR #518, opened 2024-03-25
DM-42747: Fiducial values for summary stats for comCamSim

kadrlica:
Adds fiducial values of seeing, sky brightness, and zeropoints for comCamSim in preparation for processing of Ops Rehearsal 3. Note: fiducials are not set for the `characterizeImage` task, since there is no comCamSim-specific config for this task.

### DIFF
--- a/config/comCamSim/calibrate.py
+++ b/config/comCamSim/calibrate.py
@@ -49,3 +49,6 @@ config.photoCal.match.referenceSelection.doMagLimit = True
 config.photoCal.match.referenceSelection.magLimit.fluxField = "lsst_r_flux"
 config.photoCal.match.referenceSelection.magLimit.minimum = 14.0
 config.photoCal.match.referenceSelection.magLimit.maximum = 22.0
+
+# Exposure summary stats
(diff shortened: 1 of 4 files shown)


laurenam:
> Note: fiducials are not set for the characterizeImage task, since there is no comCamSim-specific config for this task.

...is not a reason to skip adding a needed config override (i.e. you would need to add the file).  But in this case, this:
https://github.com/lsst/pipe_tasks/blob/main/python/lsst/pipe/tasks/characterizeImage.py#L173-L185
is a good reason not to include it 😉

kadrlica:
Yeah, that is a much better reason :). 

Here is the successful Jenkins run [972](https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/972/pipeline).

jrmullaney:
This looks fine to me. Just a minor thing...I suggest you expand a little on the "Plots can be found in DM-42747" comment. Perhaps "Plots, including the median values used here, can be found on JIRA ticket DM-42747"? It'll need a rebase before merging as well.

Happy to sign-off on this - no need to re-request.

jrmullaney (approved):

PR #519, opened 2024-03-28
DM-43576: ComCamSim config overrides for CalibrateImage cannot be applied

kfindeisen:
This PR fixes an incorrectly named variable in the CalibrateImage config.


parejkoj (approved):
Other than my annoyances that we apparently have multiple conventions for the config directory variable name, and that we don't have any CI for comCamSim things (neither of which are your problem!), this is fine.

PR #520, opened 2024-03-28
DM-43583: Configure flake8 to be able to lint obs_lsst configs

parejkoj:


config/.ruff.toml
```diff
@@ -0,0 +1,7 @@
+extend = "../pyproject.toml"
+# Ignore undefined "config", as config files are loaded with `eval`.
+builtins = ["config"]
+
+[pycodestyle]
```
timj, line 5:
```suggestion
[lint.pycodestyle]
```

config/.ruff.toml
```diff
@@ -0,0 +1,7 @@
...
+# Allow longer docstrings in configs, to better match the long single-lines.
+max-doc-length = 110
```
timj, line 7: It looks like there are only 10 places where the lines are slightly long. Can we fix those lines so we don't have to set this at all here? It's all python code so I don't really want the config python rules to be different to the main python code rules if we don't need them to be different.

timj (approved):
Looks okay. I see you ran black on these but without having a black configuration block in the pyproject.toml. Might be worth adding the standard black configuration in so that if this package does adopt black it won't trigger any changes in the config code (I don't think it will make a difference but it might).

Also I don't know why we are allowing long docstrings in comments. There are 10 places that need tweaking to make it pass with the standard rules.

config/.ruff.toml
parejkoj: My original intent with this was to leave longer lines as-is because we write these configs out (with comments as documentation) to the butler as `*_config` outputs, and it doesn't really make sense to have those files (which often have very long lines, sometimes >110 chars because of `config` expansion into the full task name) be forced to conform to this standard as well.

I don't care that much, and I can make this package conform to <80char comment lines, but I do think it's not necessarily a good fit for config files (which aren't exactly "code", in many ways).

parejkoj:
I didn't accept all of the black choices: it un-wrapped some of the manually-wrapped lines, which I think made them harder to read in context. I would much rather have a larger conversation about using black for config files than hang up this RFC (which is to just get our configs checked for basic errors) on that question.

config/.ruff.toml
timj: I agree that the config files that we serialize have long comments (the entire docstring of the field in one line) and long code lines (the long assert at the top) but these aren't serialized configs but are ones written manually by people aren't they?

parejkoj: Yes, these are manually written. I'd forgotten that the lines in the output files are the docstrings, so I guess that concern doesn't apply.

I still think it's a bit odd, because these configs don't feel exactly like "code", but I've changed it to have comment lines under 80chars.

PR #522, opened 2024-04-04
DM-43674: Switch comCamSim to use the LSSTCam distortion model

isullivan:


jmeyers314:
What's the long term plan here?  I anticipate we'll make more ComCamSim sims going forward, presumably without accidentally using the wrong optics.  For those we'll want the current numbers, not the ones in this PR.

(I'm fine with this as a temporary override, of course).

TallJimbo (approved):
Looks good!

To partially answer @jmeyers314 's question, I think this is a reminder that we need to eliminate code that relies on obs_lsst to get the camera object in favor of code that reads the butler camera dataset.  And then we can give the butler camera dataset real bounded validity ranges (corresponding to old and new sims) and see what falls apart.  Something surely will, so that won't be pleasant, but we need to get this sorted before we experience a real change in a real camera.

timj:
It seems odd to say that ComCamSim is using LSSTCam model and not LSSTComCam model.

PR #523, opened 2024-04-12
DM-42828: Add camera team crosstalk nonlinearity results

plazas:


python/lsst/obs/lsst/script/write_non_linear_crosstalk_coeffs.py
```diff
@@ -0,0 +1,178 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+import re
+import os
+import dateutil.parser
+import numpy as np
+from lsst.daf.butler import Butler
+from lsst.ip.isr import CrosstalkCalib
+import lsst.utils
+
+repo = "/sdf/group/rubin/repo/main"
```
czwa, line 29: This would be safer if the files were dumped from butler into the obs_lsst_data package, so this isn't location specific.

python/lsst/obs/lsst/script/write_non_linear_crosstalk_coeffs.py
```diff
@@ -0,0 +1,178 @@
...
+        crosstalk_results = butler.get(ref)
+        for j in range(nAmp):
+            target_segment = AMP2SEG[j + 1]
+            for k in range(nAmp):
+                source_segment = AMP2SEG[k + 1]
+                if j == k:
+                    c0_matrix[k, j, i] = 0.0
+                    c1_matrix[k, j, i] = 0.0
+                    continue
+                try:
```
czwa, line 117: Why does this fail that we need a try/except block?

python/lsst/obs/lsst/script/write_non_linear_crosstalk_coeffs.py
```diff
@@ -0,0 +1,178 @@
...
+    for i, ref in enumerate(data_refs):
+        crosstalk_results = butler.get(ref)
+        for j in range(nAmp):
+            target_segment = AMP2SEG[j + 1]
+            for k in range(nAmp):
+                source_segment = AMP2SEG[k + 1]
+                if j == k:
+                    c0_matrix[k, j, i] = 0.0
+                    c1_matrix[k, j, i] = 0.0
+                    continue
```
czwa, line 116: I don't know that it's essential, but these should set the `coeffValid` entry to `False`.  If we had made these calibrations with `cp_pipe`, they would be set that way (as a coefficient of zero should be smaller than its error), and it's good to be consistent.

czwa (commented):
I think the coeffValid comment is the only thing that needs to be fixed.  Do you have an idea of how large/how difficult it would be to shift the things being retrieved from butler into source files in the `_data` package would be?  I would prefer they be there, so this script is portable (and safe against someone accidentally overwriting the collections).

timj:
The standard approach is to create the ECSV files and store them in obs_lsst_data and then they become standard curated calibrations that can be loaded into any butler. The usual thing is that the ECSV files are read and then butler.put turns them into FITS to let them be read as fast as possible, but the files are small enough that storing them as text in git makes it easier.

plazas:
> The standard approach is to create the ECSV files and store them in obs_lsst_data and then they become standard curated calibrations that can be loaded into any butler. The usual thing is that the ECSV files are read and then butler.put turns them into FITS to let them be read as fast as possible, but the files are small enough that storing them as text in git makes it easier.

Thanks. This script is already getting the coefficients from the Butler and saving them as `ecsv` files in `obs_lsst_data` (there's another PR in there with this same ticket number). Is that sufficient in this case for this ticket, or should the `butler.put` step to turn them into fits files also be implemented?

timj:
If this is a standard IsrCalib class (which it should be) then it should already be storing as FITS in the butler when it's read from obs_lsst_data.

python/lsst/obs/lsst/script/write_non_linear_crosstalk_coeffs.py
plazas: For the diagonal coefficients (same amp), do we currently set them as invalid?

czwa: We should be: https://github.com/lsst/cp_pipe/blob/main/python/lsst/cp/pipe/measureCrosstalk.py#L642

czwa:
My concern was that the source data would be lost, leaving us no way to regenerate these files if needed.  I think ensuring that the script uses the RUN collection instead of the CHAINED collection ensures that we can find things.  A lot of the other curated calibrations have no clear provenance, so I've pushed for the sources of those products to be included in the _data package.  I think we're safer with things in Butler.

timj:
So long as write-curated-calibrations is using obs_lsst_data and not trying to find a specific butler that's fine. I assume there is a readme in obs_lsst_data that says where the data came from.

czwa:
> So long as write-curated-calibrations is using obs_lsst_data and not trying to find a specific butler that's fine. I assume there is a readme in obs_lsst_data that says where the data came from. ie the scripts that created the ECSV files from butler are part of the provenance but not something used regularly.

Yes, write-curated-calibrations is doing the standard thing (pull from obs_lsst_data, and write to the target butler).  My concern was prompted by things like the transmission curves being based on an xls spreadsheet (where that one file and some email quotes are the only provenance we have).  Pointing at the butler RUN collections should be stable and reproducible, so I think my concerns were inflated.

python/lsst/obs/lsst/script/write_non_linear_crosstalk_coeffs.py
plazas: From Adam: "when we take the images there are only a subset that have actual crosstalk measurements (since we need to readout at least a REB worth of images). So when I run the analysis there are some empty crosstalk datasets where there was no source detected for that specific amp pair."

python/lsst/obs/lsst/script/write_non_linear_crosstalk_coeffs.py
czwa: Commenting here that I'm not worried about this anymore.

czwa (approved):

PR #532, opened 2024-06-04
DM-43568: Add support for can_see_sky metadata translation

timj:
* No test headers included yet. Need an on-sky LATISS observation and a flat to show it works properly.
* Relies on SHUTTIME being 0 for dark and bias.
* Assumes flats will indicate vignetting.
* Currently if VIGN_MIN is not in the header the calculation falls back to using the observation type. We might want to disable that for non-simulated cameras after comcam on-sky date.


tests/headers/latiss-AT_O_20240624_000106_R00_S00-patched.yaml
```diff
@@ -0,0 +1,269 @@
...
+RADESYS: ICRS
+FOCUSZ: 0.742219865322113
+OBJECT: HD 131586
+INSTPORT: 2
+ATM3PORT: null
+DOMEAZ: 358.96
+SHUTLOWR: 0.0
+SHUTUPPR: 100.0
+VIGNETTE: 'NO'
+VIGN_MIN: 'FULLY'
```
ktlim, line 84: Isn't this impossible in real life?

python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -969,6 +969,7 @@ def to_has_simulated_content(self):
         # true value implies that something in the data is simulated.
         for k, v in self._header.items():
             if "SIMULATE" in k and v:
+                self._used_these_cards(k)
```
ktlim, line 972: You don't need to mark other `SIMULATE` cards as used, even if they're false or after the first true one?

ktlim (approved):

tests/headers/latiss-AT_O_20240624_000106_R00_S00-patched.yaml
timj: Correct. But I'm testing it to make sure I am really reading the header and not falling back to observation type. If it makes you feel better I can change the Vignette header as well.

python/lsst/obs/lsst/translators/lsst.py
timj: This is the card that is changing the state but more importantly I'm trying to avoid reading every card looking for simulate values. This whole approach is a bit inefficient. In my previous telescope the header service looked at all the simulate headers to create a new derived simulate header so the readers only had to look in one known place.

tests/headers/latiss-AT_O_20240624_000106_R00_S00-patched.yaml
timj: I made VIGNETTE consistent.

PR #535, opened 2024-06-06
DM-44666: Update fgcmFitCycle LATISS physicalFilterMap with limited list of input filters.

erykoff:
This can be reverted to the full filter list with DM-44705 which will redefine the plot connections.

### DIFF
--- a/config/latiss/fgcmFitCycle.py
+++ b/config/latiss/fgcmFitCycle.py
@@ -1,12 +1,28 @@
 import lsst.fgcmcal as fgcmcal
 
+# We use our own filter to band mapping to limit to only the
+# filters used in the photometric survey which may be input
+# to fgcmcal. This override will be less important after
+# TODO: DM-44705.
+physical_to_band = {
+    "SDSSg_65mm~empty": "g",
+    "SDSSr_65mm~empty": "r",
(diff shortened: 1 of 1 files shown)


config/latiss/fgcmFitCycle.py
```diff
@@ -1,12 +1,28 @@
...
+# to fgcmcal. This override will be less important after
+# TODO: DM-44705.
+physical_to_band = {
+    "SDSSg_65mm~empty": "g",
+    "SDSSr_65mm~empty": "r",
+    "SDSSi_65mm~empty": "i",
+    "empty~SDSSi_65mm": "i",
+    "SDSSz_65mm~empty": "z",
+    "SDSSy_65mm~empty": "y",
+    "empty~SDSSy_65mm": "y",
```
leeskelvin, line 14: There are only `empty~SDSSx` variants for the `i` and `y` bands?

leeskelvin (approved):

config/latiss/fgcmFitCycle.py
erykoff: Yup, those are the only two that have been moved to the other filter wheel.

PR #537, opened 2024-06-10
DM-44749: Add coadd config overrides for LATISS

arunkannawadi:


config/latiss/compareWarpAssembleCoadd.py
```diff
@@ -0,0 +1,27 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+
+import os.path
+
+# Load configs from base assembleCoadd
+config.load(os.path.join(os.path.dirname(__file__), "assembleCoadd.py"))
```
leeskelvin, line 26: This should be `assembleCoadd`, rather than `coaddBase`, right?

leeskelvin (approved):

config/latiss/compareWarpAssembleCoadd.py
arunkannawadi: Correct. The config overriding follows the same order as class inheritance. `assembleCoadd` overrides the default configs with those specified in `coaddBase` before any further overriding, and similarly `compareWarpAssembleCoadd` overrides with those specified in `assembleCoadd`.

arunkannawadi: Here is an example: https://github.com/lsst/obs_lsst/blob/main/config/compareWarpAssembleCoadd.py

PR #545, opened 2024-07-23
DM-45378: Add script that can be run to average lsstcam ITL crosstalks.

erykoff:


python/lsst/obs/lsst/script/write_comcam_estimated_non_linear_crosstalk_coeffs.py
```diff
@@ -0,0 +1,122 @@
...
+c1_error_matrix[np.diag_indices(n_amp)] = 0.0
+
+for det in comcam_camera:
+    name = det.getName()
+    raft, sensor = name.split("_")
+
+    comcam_crosstalk = CrosstalkCalib(nAmp=n_amp)
+    comcam_crosstalk.coeffs = c0_matrix
+    comcam_crosstalk.coeffErr = c0_error_matrix
+    comcam_crosstalk.coeffsSqr = c1_matrix
```
czwa, line 96: I went back and forth whether including these is the right thing to do or not, eventually settling on "it probably doesn't hurt."

czwa (approved):

python/lsst/obs/lsst/script/write_comcam_estimated_non_linear_crosstalk_coeffs.py
erykoff: Aaron Roodman explicitly suggested using the non-linear terms, so that's what I've done.

python/lsst/obs/lsst/script/write_comcam_estimated_non_linear_crosstalk_coeffs.py
```diff
@@ -0,0 +1,122 @@
...
+    comcam_crosstalk.coeffErrSqr = c1_error_matrix
+    comcam_crosstalk.ampGainRatios = gain_ratio_matrix
+    comcam_crosstalk.crosstalkRatiosUnits = "electron"
+    comcam_crosstalk.coeffValid = np.ones((n_amp, n_amp), dtype=np.bool_)
+    comcam_crosstalk.hasCrosstalk = True
+
+    out_path = os.path.join(data_root, "comCam", "crosstalk", name.lower())
```
timj, line 103: In theory if you have the instrument class you can do

```suggestion
    out_path = os.path.join(instr.getObsDataPackageDir(), instr.policyName, "crosstalk", name.lower())
```
and guarantee consistency with what the instrument class is going to do in write-curated-calibs (which is the wrong name and should have been ingest-curated-calibs). There is a private method in lsst.obs.base.Instrument that could be public

PR #546, opened 2024-07-24
DM-44908-hotfix: Fix CalibrateImage LATISS psf fitting overrides to be fully consistent.

erykoff:
Prior to this fix the psf code was normalizing to 12 pixel aperture fluxes which only accidentally worked.  When I fixed the aperture radii to use the preferred LATISS 35 pixels that turned out to break the psf fitting.


laurenam:
Arghh...I had already noted the need for these on DM-45072 (in [this comment](https://rubinobs.atlassian.net/browse/DM-45072?focusedCommentId=517982)).  Sorry for not shouting louder (I'm putting it in review to you now 🤪)

TallJimbo (approved):

PR #547, opened 2024-07-30
DM-22302: Prefer -BEG headers over -OBS headers for datetime_begin

timj:


python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -529,8 +529,18 @@ def to_location(self):
...
     def to_datetime_begin(self):
         # Docstring will be inherited. Property defined in properties.py
-        self._used_these_cards("MJD-OBS")
-        return Time(self._header["MJD-OBS"], scale="tai", format="mjd")
+        # Prefer -BEG over -OBS. Let it fail inside Time if no headers
+        # can be found.
+        date_key = "MJD-BEG"
+        date_fmt = "mjd"
+        for k in ("MJD-BEG", "DATE-BEG", "MJD-OBS", "DATE-OBS"):
+            if self.is_key_ok(k):
```
czwa, line 537: I'm not sure I understand how this interacts with the default values.  If none of these keys exist, it will still try to use `MJD-BEG` in the return value.  Won't that cause a `KeyError` on L543?

timj: Yes. Which will fail the translation. The translation has to blow up anyhow if there aren't any keys available. I was mostly saving myself the trouble of writing my own exception message.

PR #548, opened 2024-08-05
DM-39168: Turn off proper motion correction

cmsaunders:


parejkoj (changes requested):
Please put a comment above each of these lines saying why you are disabling PM corrections. I think I know the answer for comCamSim and imsim, but I don't for LATISS, and a general developer probably won't know the history of any of those.

parejkoj (approved):

PR #549, opened 2024-08-07
DM-45651: Write configs for ComCam and LSSTCam

mfisherlevine:


kadrlica:
@mfisherlevine updated effTime fiducials for ComCam can be found here:
https://github.com/lsst/obs_lsst/blob/tickets/DM-45333/config/comCam/computeExposureSummaryStats.py

mfisherlevine:
> @mfisherlevine updated effTime fiducials for ComCam can be found here: https://github.com/lsst/obs_lsst/blob/tickets/DM-45333/config/comCam/computeExposureSummaryStats.py
> 
> It'd be useful to also grab the updated comment at the top of that file.

Done - I grabbed that wholesale, and also added text from our conversation about how the LSSTCam ones are currently an estimate and it's not obvious how to do that across the chips.

config/imsim/calibrate.py
```diff
@@ -44,6 +44,7 @@
 
 # No color terms in simulation.
 config.photoCal.applyColorTerms = False
+config.photoCal.photoCatName = 'ps1_pv3_3pi_20170110'
```
yalsayyad, line 47: Oh my. How did this ever work?

config/refCatObjectPhotometricAnalysisTask.py
```diff
@@ -0,0 +1,4 @@
+"""Overrides for RefCatObjectPhotometricAnalysisTask"""
```
yalsayyad, line 1: This whole file should be in the ComCamSim directory

yalsayyad (approved):
Let's file another ticket to move what we can into the parent config/ directory. Like what we did for calibrate.. but for measureCoaddSources, etc...

config/imsim/calibrate.py
mfisherlevine: Are you saying this needs changing, or are you just noticing this fixes a pre-existing problem (there's a few places that were actually like that). I don't actually know catalog imsim uses for simulation...

yalsayyad: panstarrs for imsim is definitely wrong, and it looks like it's only resolving to the right config because we override them in drp_pipe: https://github.com/lsst/drp_pipe/blob/main/pipelines/_ingredients/LSSTCam-imSim/DRP.yaml#L42

mfisherlevine: Does [this commit](https://github.com/lsst/obs_lsst/pull/549/commits/874f64a4288b11f78dd0a6c558b675cca2221810) put it how you'd like it then?

PR #550, opened 2024-08-19
DM-45391: Add scripts to make baseline ComCam curated calibrations. 

erykoff:
These are scripts to (a) average the ITL sensor QE curves; (b) make filter and optics transmission curves based on the official throughputs v1.9.

### DIFF
--- a/python/lsst/obs/lsst/script/write_comcam_estimated_transmission_sensor.py
+++ b/python/lsst/obs/lsst/script/write_comcam_estimated_transmission_sensor.py
@@ -1,0 +1,133 @@
+# This file is part of obs_lsst.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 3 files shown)


python/lsst/obs/lsst/script/write_comcam_estimated_transmission_sensor.py
```diff
@@ -0,0 +1,126 @@
...
+
+data_root = lsst.utils.getPackageDir("obs_lsst_data")
+
+valid_start = "1970-01-01T00:00:00"
+valid_date = dateutil.parser.parse(valid_start)
+datestr = "".join(re.split(r"[:-]", valid_date.isoformat()))
+
+# Get the transmission_sensor filenames of every ITL detector.
+transmission_sensor_files = []
+for det in lsst_camera:
```
jchiang87, line 45: Is it worth putting a comment somewhere in this file stating why the LSSTCam ITL curves are being used? I had to look at the corresponding README in `obs_lsst_data` to make sure I understood what's happening here.

python/lsst/obs/lsst/script/write_comcam_estimated_transmission_sensor.py
```diff
@@ -0,0 +1,126 @@
...
+    os.makedirs(out_path, exist_ok=True)
+    out_file = os.path.join(out_path, f"{datestr}.ecsv")
+
+    curve_table.meta.update(
+        {
+            "CALIBDATE": valid_start,
+            "INSTRUME": "ComCam",
+            "OBSTYPE": "transmission_sensor",
+            "TYPE": "transmission_sensor",
+            "DETECTOR": det_num,
```
jchiang87, line 113: Should there be a `"SOURCE"` metadata entry somehow indicating that the median of the LSSTCam ITL CCDs was used?

jchiang87 (changes requested):
I think it would be useful to document explicitly that the LSSTCam ITL CCDs are being used for the estimated ComCam QE curves, otherwise all looks good.

erykoff:
@jchiang87 Thanks!  I think this is ready now.

jchiang87 (approved):
Thanks for the updates!  Looks good.

PR #551, opened 2024-08-22
DM-45886: Reconfigure refcats while we're working with sims

mfisherlevine:


config/lsstCam/calibrate.py
```diff
@@ -2,3 +2,6 @@
 
 obsConfigDir = os.path.join(os.path.dirname(__file__))
 config.photoRefObjLoader.load(os.path.join(obsConfigDir, "filterMap.py"))
+
+config.photoCal.photoCatName = 'gaia_dr3_20230707'
+config.connections.photoRefCat = 'ps1_pv3_3pi_20170110'
```
erykoff, line 7: I don't understand gaia as the photoCatName?

mfisherlevine: Discussed OOB, moving to a user branch for Guillem.

PR #564, opened 2024-10-16
DM-46918 update config/ComCam to use the_monster as a refcat

psferguson:
Here we update the configuration for comcam to use the_monster as a refcat


parejkoj:
I'll note that we'll probably want to change this to be the task defaults for all tasks that take a refcat. Starting it here for just ComCam seems reasonable, but once we're through commissioning (or once we've demonstrated it works), we should make it the default and remove these overrides.

erykoff (approved):

PR #565, opened 2024-10-22
DM-47034: Convert effTime fiducials to electrons

kadrlica:
ComCam, ComCamSim and lsstCam will produce images in units of electrons. This means that the sky background and zeropoint values should be updated to be in units of electrons. This is actually the native unit of SMTN-002, so in many ways this is cleaner than what was being done before (which involved estimating and applying the gain).


kadrlica:
Jenkins run [#2869](https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/2869/pipeline)

mfisherlevine (approved):

PR #566, opened 2024-10-22
DM-47038: Add pinhole, none, OG590_2_5mm filters for ComCam.

ktlim:


python/lsst/obs/lsst/filters.py
```diff
@@ -443,14 +443,15 @@ def addFilter(filter_dict, band, physical_filter):
...
     physical_filter = f"{band}_{sn[3:]}"
-    lsstCamFilter = [f for f in LsstCamFiltersBaseline if f.band == band][0]
-
     addFilter(ComCamFilters_dict, band, physical_filter)
 
 
 ComCamFilters = [
     FilterDefinition(band="white", physical_filter="empty"),
     FilterDefinition(band="unknown", physical_filter="unknown"),
+    FilterDefinition(band="white", physical_filter="none"),
```
timj, line 452: Don't we use `NoFilterCollection` for this definition?

python/lsst/obs/lsst/filters.py
```diff
@@ -443,14 +443,15 @@ def addFilter(filter_dict, band, physical_filter):
...
+    FilterDefinition(band="white", physical_filter="pinhole"),
+    FilterDefinition(physical_filter="OG590_2_5mm"),
```
timj, line 454: No band? What type of filter is this?

ktlim: It's some kind of narrowband filter.

timj: Looks like, from HSC, we would name the band OG590 then or something similar. (depending on what OG means)

python/lsst/obs/lsst/filters.py
ktlim: Ah, yes, that looks better, changed to using it.

timj (approved):

PR #568, opened 2024-10-24
DM-46188: changes to get LSST pipelines passing smoke tests in drp_pipe

TallJimbo:


erykoff (changes requested):
We changed LATISS to the monster from ATLAS recently.

TallJimbo:
> We changed LATISS to the monster from ATLAS recently.

@mfisherlevine seems to have done that and then undone that for photometry on the very latest commit on `main` (as of this writing):

https://github.com/lsst/obs_lsst/commit/e3e606bae89818bd3ff4ec56999f23786f264f6f#diff-8fbcd878082ed6a16f8d187f42ff8e31e44a015dbfa7cfb4b047d8713101876dL12

erykoff:
What?

erykoff (commented):
I think that https://github.com/lsst/obs_lsst/blob/main/config/calibrate.py astrometry refcat should be updated as well to use the monster for astrometry.

TallJimbo:
I started on this but it ballooned out of control, and I think I need to limit the scope of this ticket to the minimum needed to get the tests running with the pipelines configured more or less as they were.  I've saved my progress on branches of DM-46454 and I will return to that work tomorrow, on that ticket or a new one.

erykoff (commented):
I think the monster needs to be set as the astrometry refcat in `calibrate.py`

erykoff (approved):

PR #569, opened 2024-10-27
DM-47165: Update astrom/photoCal for early ComCam data.

ctslater:


config/comCam/filterMap.py
```diff
@@ -1,18 +1,12 @@
...
-    ("i_06", "monster_SynthLSST_i"),
-    ("z_02", "monster_SynthLSST_z"),
-    ("z_03", "monster_SynthLSST_z"),
-    ("y_04", "monster_SynthLSST_y"),
+    ("u", "monster_SynthLSST_u"),
+    ("g", "monster_SynthLSST_g"),
+    ("r", "monster_SynthLSST_r"),
+    ("i", "monster_SynthLSST_i"),
+    ("z", "monster_SynthLSST_z"),
+    ("y", "monster_SynthLSST_y"),
```
erykoff, line 10: I think for the time being we want to have both here, or we can fix this when we fix the photoCal code which is fundamentally broken because *it uses band if colorTerms are off and uses filter if colorTerms are on and that's insane*.  I think the RFC was about cleaning that up, but it's definitely wrong at the moment (by definition).  If we want to calibrate empty/white the lines should be:
```
    ("empty", "monster_SynthLSST_r"),
    ("white", "monster_SynthLSST_r"),
```
(One for filter, the other for band).

erykoff: And if we are never photocalibrating empty those can be left off.

erykoff (approved):

config/comCam/calibrate.py
```diff
@@ -3,3 +3,9 @@
 obsConfigDir = os.path.join(os.path.dirname(__file__))
 config.photoRefObjLoader.load(os.path.join(obsConfigDir, "filterMap.py"))
 config.connections.astromRefCat = "the_monster_20240904"
+
+# Loosen maxOffset to account for early pointing model inaccuracy.
+config.astrometry.matcher.maxOffsetPix=500
```
laurenam, line 8: Did you note the comment about also updating `LoadReferenceObjectsConfig.pixelMargin` in the docstring for that config:
```
    maxOffsetPix = pexConfig.RangeField(
        doc="Maximum allowed shift of WCS, due to matching (pixel). "
            "When changing this value, the "
            "LoadReferenceObjectsConfig.pixelMargin should also be updated.",
        dtype=int,
        default=250,
        max=4000,
    )
```
Also, I think the padding in for the visit definition may need to be considered if things are really bad:
https://github.com/lsst/obs_base/blob/2b7fa98c38048493aa6c69f9e8fa92a242ee93f2/python/lsst/obs/base/defineVisits.py#L276-L289

ctslater: Thank you! Added `config.astromRefObjLoader.pixelMargin=500` for consistency,

config/comCam/filterMap.py
ctslater: Ok thanks, I'll add all-of-the-mappings for now and I left a note with your explanation why and the RFC.

PR #580, opened 2024-11-18
DM-47636: Switch CROSSTALK masking to use subtrahend masking

czwa:
Enable subtrahend masking.


erykoff (approved):
This is fine, but we probably should at some point before LSSTCam figure out what should go in the parent `isrLSST.py` (which doesn't exist yet so is out of scope for this ticket).

PR #584, opened 2024-11-26
DM-47365: new manual defects for LSSTComCam

aferte:


python/lsst/obs/lsst/script/write_comcam_manual_defects.py
```diff
@@ -0,0 +1,46 @@
...
+
+cc = ol.LsstComCam
+camera = cc.getCamera()
+for det in camera:
+    detId = det.getId()
+    detName = det.getName()
+
+    # This is only valid for LSSTComCam which has a single raft
+    manual_defects = \
+        Defects.readText('obs_lsst_data/comCam/manual_defects/r22_s'+detName[5:]+'/20241120T000000.ecsv')
```
czwa, line 42: This may be redundant with the changes in `obs_base` that make these curated calibrations that are ingested into the butler with `butler write-curated-calibrations`.  The files are small, and doing it this way means we can generate (and regenerate) the defects prior this this merging, so I don't think it's that big of a deal.
However, I think this would be better getting the path dynamically, like we do in other scripts like this (https://github.com/lsst/obs_lsst/blob/main/python/lsst/obs/lsst/script/write_latiss_crosstalk_coeffs.py#L42 for reference).  `getPackageDir` and `os.join` make it safer to construct paths like this, as it will use whichever `obs_lsst_data` source is set up.

czwa (approved):
One request that will just make this more consistent with some of our other scripts.

PR #588, opened 2024-12-05
DM-47919: Update fgcmcal configs for ugrizy high latitude ComCam science observations

erykoff:


laurenam (approved):
Commit message is too long.  Could leave out ugrizy and add it as a short description?

PR #590, opened 2024-12-09
Increase spacial order interpolation for PSF/Piff

PFLeget:
Based on first iteration if ComCam I can see that increasing spatial interpolation of PSF parameters from order 2 to order 3 it removes most of spatial structures in second moment of PSF second moment residuals.


fred3m (approved):
LGTM

PR #593, opened 2024-12-18
DM-48174: Update ComCam manual defect script and add new manual defects

erykoff:


python/lsst/obs/lsst/script/write_comcam_manual_defects.py
```diff
@@ -18,36 +18,107 @@
...
+
+    # Define the manual defects from DM-47365 and DM-48174.
+    if det.getId() == 0:
+        # Ardenti Negans Imperfecta (incomplete glowing negative column).
+        box_xywh.extend(
+            (
+                (680, 2000, 11, 966),
+            ),
+        )
+    if det.getId() == 1:
```
czwa, line 60: As noted in the discussions earlier, this could/should be an `elif`.  Maybe it'd be better as a `match/case` block, but I have no strong opinions about whether that's faster or more legible, so I mention it only for completeness.

czwa (approved):

PR #599, opened 2025-01-27
Final set of AOS comcam on-sky group_id corrections.

jbkalmbach:
We discovered that the initial corrections made to the early AOS ComCam group ids only covered the last set of triplets in a closed loop and did not correct previous iterations. Making this last correction to all triplets that have the same group number will get us at least 10% more usable comcam data for AOS analysis.


timj (approved):
Looks great. You are making me ponder my life choices in this implementation mechanism though 😄

PR #603, opened 2025-02-14
DM-48887: Enable zeroth order interp if not enough stars.

PFLeget:
More context in https://rubinobs.atlassian.net/browse/DM-48887.


erykoff (approved):
This is fine, but I think it could be made the overall default making this unnecessary.  (See other PR comments)

PFLeget:
I guess for the moment as commented in the other PR, I would like to have it as optional until I make something more clean and tractable.

PR #604, opened 2025-02-14
DM-47181: Add gbdes config file for LSSTComCam

PFLeget:
In the configurations of gbdes I am just enabling Proper Motion fitting right now.


PFLeget:
I will close this PR as I implemented this change in DRP pipe instead of here as suggested.

PR #608, opened 2025-02-25
DM-49166: Add ComCam overrides for make(PsfMatched|Direct)Warp

arunkannawadi:


config/comCamSim/makeDirectWarp.py
```diff
@@ -0,0 +1,34 @@
...
+# These thresholds can be tightened for the simulated data compared to the
+# task defaults (which were conditioned on HSC data).  These have been
+# chosen based on the OR3 dataset reduced with w_2024_19 pipeline on
+# DM-37952 (see ticket for figures).
+config.doSelectPreWarp = True
+config.select.maxEllipResidual = 0.004
+config.select.maxScaledSizeScatter = 0.014
+config.select.maxPsfTraceRadiusDelta = 0.091
+config.select.maxPsfApFluxDelta = 0.047
+config.select.maxPsfApCorrSigmaScaledDelta = 0.041
```
TallJimbo, line 34: Are these duplicated with the makeWarp configs for this instrument, then, rather than shared as you did for comCam?

TallJimbo (approved):

config/comCamSim/makeDirectWarp.py
arunkannawadi: I'm a bit conflicted about loading the configs of one task to another, since there are configs that don't match up and depending on what overrides end up where, it could break things. Given the current overrides and pending deprecation, it's not an issue now and can homogenize it either way.

yalsayyad (approved):

config/comCam/makeWarp.py
```diff
@@ -22,18 +22,10 @@
...
+import os.path
 
-# PSF-matching configs are in units of pix and specific to skymap pixel scale.
+configDir = os.path.dirname(__file__)
+config.load(os.path.join(configDir, "makeDirectWarp.py"))
 
-# DM-47171: 9.0 (1.8 arcsec) corresponds to the ~98.5% percentile of the
-# PSF FWHM distribution in the 2nd and 3rd weeks of ComCam science observations
-# selectDeepCoaddVisits is currently set the the default 2.0 arcsec
+# makePsfMatchedWarp configs cannot be loaded in the same way in general.
```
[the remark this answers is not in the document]
arunkannawadi: No, the reason is if someone added a line like `config.warpAndPsfMatch.psfMatch.kernel["AL"].kernelSize = 29`, loading that here would result in an error. Given that the classes don't obey an inheritance relationship, it would not be a good idea to load `makePsfMatchedWarp` config file into `makeWarp`. I will go ahead with the comment about keeping this in sync.

PR #609, opened 2025-02-26
DM-49155: Turn on illumination corrections for LSSTComCam.

erykoff:


config/comCam/makeDirectWarp.py
```diff
@@ -31,3 +31,5 @@
 config.select.maxPsfTraceRadiusDelta = 4.4
 config.select.maxPsfApFluxDelta = 1.6
 config.select.maxPsfApCorrSigmaScaledDelta = 0.13
+
+config.doApplyFlatBackgroundRatio = True
```
yalsayyad, line 35: Does this config parameter exist in makeWarp?  And did you do your Jenkins run before Arun merged the update? This might need some surgery.

yalsayyad (approved):

config/comCam/makeDirectWarp.py
erykoff: This is just applied to `makeDirectWarp.py`, and this was committed after Arun merged his update.  I don't know why there would be a `makeWarp.py` anyway.

Meanwhile, if this isn't set correctly then the pipeline contracts would fail in drp_pipe because it has to be consistent with the other `obs_lsst` illumination correction overrides.

PR #615, opened 2025-03-25
DM-49673: Add per-band spatialOrder config option in piff and add a minNPsfStar criterion for detector rejection

laurenam:


config/comCam/finalize_characterization.py
```diff
@@ -1,7 +1,3 @@
-# DM-48093: Order 2 does not looks to be enough
-# for spatial interpolation of the PSF. Order 4
-# remove most of spatial structure.
-config.psf_determiner['piff'].spatialOrder = 4
```
PFLeget, line 4: I would be in favor to have explicitly the per filter setup here, with the tickets as a reference.

PFLeget (approved):

PR #621, opened 2025-04-02
DM-49679: Add script to generate LSSTCam pseudoflats.

erykoff:


czwa (approved):
Took a bit to follow all the steps, but this looks reasonable.

PR #627, opened 2025-04-09
DM-50002: Account for the sequence number now being independent of controller

timj:
This allows the exposure ID to no longer have an offset applied for each controller.


ktlim (approved):
Code and test looks OK.  As I read it, the date needs to be after deployment at all ingest sites, correct?

timj:
> the date needs to be after deployment at all ingest sites, correct?

Yes. I've moved the date out to the 17th April to allow some time for the migration to new ingest weekly.

PR #628, opened 2025-04-09
DM-49997: Drop controller from dimension packing by default.

TallJimbo:
Old behavior can be recovered by calling use_controllers() on the config object, which is what some of the tests do and how the compute_detector_exposure_id implementation on the translator now works.


TallJimbo:
Wrong ticket number.

TallJimbo:
No, right ticket number.

TallJimbo:
No, I was going to do that on DM-49134 instead.  But now that you've made me think about it, it really belongs here instead.  `meas_base` PR incoming!

PR #631, opened 2025-04-10
DM-44113: Test and update electrometer ingest for FITS files as currently written.

czwa:
The main blockers on this is now how reliable the blocks/scheduler/etc are at generating GROUPID values.  I think this is ready for review, with the intent that we'll run this manually until we know what that reliability is.

### DIFF
--- a/python/lsst/obs/lsst/_ingestPhotodiode.py
+++ b/python/lsst/obs/lsst/_ingestPhotodiode.py
@@ -38,7 +38,7 @@ from lsst.pipe.base import Task
 from lsst.resources import ResourcePath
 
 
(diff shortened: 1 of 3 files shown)


erykoff (approved):
I think this is fine, but does there need to be a check on the group string to make sure it follows some pattern (some non-whitespace, then #, then a number)?  What happens on the ones that have bad group numbers?

timj:
In theory we can add header correction patch files for the bad group ID values and then reingest.

czwa:
I've added code to validate the group, and skip the ingest if it doesn't match `{initial_group}#{unique_identifier}`.

erykoff (approved):

erykoff (approved):

PR #632, opened 2025-04-11
DM-50063: Fix the max_exposure_id value to be historical value

timj:
Rather than derive from controller code we have to lock it into place to prevent churn in the dimension record definition.


srp3rd (approved):
looks fine.   No changes requested.

PR #634, opened 2025-04-11
DM-50076: Allow LSSTCam on sky to be indicated by special TSTAND value

timj:
The code assumed that no TSTAND meant that the instrument was on sky, but we ended up taking on-telescope data with TSTAND = TMAMCv1_SUM so we have to look for strings like that.


mareuter (approved):
Looks good to me.

PR #635, opened 2025-04-11
Assume that "is on mountain" is always true for LSSTCam from March 2025

timj:
Even if it is on a test stand.


mareuter (approved):
Looks good to me.

PR #637, opened 2025-04-12
DM-49955: Initial LSSTCam NV and DRP pipelines

yalsayyad:


config/lsstCam/calibrateImage.py
```diff
@@ -23,12 +23,20 @@
 
 config.compute_summary_stats.load(os.path.join(config_dir, "computeExposureSummaryStats.py"))
 
+# Loosen maxOffset to account for early pointing model inaccuracy.
+config.astrometry.matcher.maxOffsetPix=1500
+config.astrometry_ref_loader.pixelMargin=1500
```
TallJimbo, line 28: The default padding for visits is just 250 ([here](https://github.com/lsst/obs_base/blob/f063a37d1876648ecbe73d35a02f3c5424e9150b/python/lsst/obs/base/defineVisits.py#L279-L289)), and I don't see anything changing that in `obs_lsst`, so if we really are off by this much, we won't be grabbing the full set of needed refcat shards in QG.  This is something we can also override in configs (`defineVisits.py`):
```
config.computeVisitRegions["single-raw-wcs"].padding = 1500
```
but we'd need that to be in the stack used by the auto-ingesters that actually run `butler define-visits`.  But I'm pretty sure we never modified that for ComCam, either, and we were fine.

TallJimbo (approved):

config/lsstCam/calibrateImage.py
yalsayyad: Yeah, we never modified the computeVisitRegions from ComCam. If I remember right, this config  meant that if you get lucky with the edge of a shard it'd use it.  I was a little surprised we never removed this from the comCam pipelines, but I guess it wasn't doing any harm which is why we didn't touch it.

PR #640, opened 2025-04-14
DM-50167: Look at TRACKSYS header and fix negative elevations

timj:
Needs lsst/astro_metadata_translator#82


python/lsst/obs/lsst/translators/lsst.py
```diff
@@ -752,6 +752,10 @@ def to_tracking_radec(self):
                 radec = astropy.coordinates.SkyCoord(altaz.transform_to(astropy.coordinates.ICRS()),
                                                      obstime=altaz.obstime,
                                                      location=altaz.location)
+            elif self.are_keys_ok(["TRACKSYS"]) and self._header["TRACKSYS"] != "RADEC":
```
timj, line 755: We also have the option, in addition to looking at TRACKSYS, to declare that flat/dark/bias should never report a tracking ra/dec even if TRACKSYS says it is tracking.

PR #641, opened 2025-04-18
DM-50338: LsstBaseTranslator.compute_exposure_id breaks with controller=None

kfindeisen:
This PR fixes a date-specific bug where `LsstBaseTranslator.compute_exposure_id` can't handle `controller=None`, despite this being the default.


timj (approved):
Thanks for adding tests.

PR #642, opened 2025-04-21
DM-50380: Remove ROTPA correction from new data and apply to old

timj:


python/lsst/obs/lsst/translators/lsstCam.py
```diff
@@ -118,13 +116,25 @@ def fix_header(cls, header, instrument, obsid, filename=None):
...
             header["FILTER"] = "ph_5"
             modified = True
 
+        # For first ~ week of observing the ROTPA in the header was the ComCam
+        # value and needed to be adjusted by 90 degrees to match LSSTCam.
+        # Fixed for day_obs 20250423.
+        rotpa_fixed_on_day = 20250423
+        if i_day_obs and i_day_obs > 20250301 and i_day_obs < rotpa_fixed_on_day:
+            if rotpa := header.get("ROTPA"):
+                header["ROTPA"] = rotpa - 90.0
```
timj, line 134: Not normalized so this will lead to the exposure dimension records changing on re-ingest because previously we corrected and wrapped 0 to 360.

erykoff (approved):
I hope this is correct!

PR #643, opened 2025-04-21
DM-45569: Add color fit for PSF fit in ComCam.

PFLeget:


config/comCam/finalize_characterization.py
```diff
@@ -4,14 +4,24 @@
...
-    "g": 4,
-    "r": 4,
-    "i": 4,
-    "z": 4,
-    "y": 4,
+    "g": 3,
+    "r": 3,
+    "i": 3,
+    "z": 3,
+    "y": 3,
```
arunkannawadi, line 11: Was this change discussed somewhere on Slack/Jira?

PFLeget: I lowered the polynomial interpolation as I was adding one more component with the color interpolation. They did something similar in DES. I did that also for speed and to avoid having not enough stars to constrain the model.

PR #649, opened 2025-04-29
DM-50529: Use None for can_see_sky if unknown

timj:
For consistency with older data sets this also required that lab data always return False since we now know when ComCam and LSSTCam arrived on the mountain.


python/lsst/obs/lsst/translators/lsstCam.py
```diff
@@ -93,6 +90,12 @@ class LsstCamTranslator(LsstBaseTranslator):
     # to summit day_offset (12 hours).
     _CAMERA_SHIP_DATE = 202405
 
+    # Date we know camera is in Chile and potentially taking on-sky data.
+    _CAMERA_ON_MOUNTAIN_DATE = astropy.time.Time("2025-03-01T00:00", format="isot", scale="utc")
```
ktlim, line 94: It was on the mountain well before this, but it was stuck in the lab.  Still, maybe "ON_TELESCOPE_DATE" would be a better phrasing?

ktlim (approved):

PR #654, opened 2025-05-22
DM-51052: Fix VIGN_MIN for LSSTCam headers on 20250518.

erykoff:


erykoff:
I have not checked that since I don't know what the next steps are.

PR #658, opened 2025-06-05
DM-51274: Add script to put in standard passbands into dp1 repo.

erykoff:


taranu (commented):
Looks fine. I see you've already run this, so is this for posterity's sake? If it's potentially meant to be run again, would the input collections need to be updated if/when they get moved off embargo?

PR #662, opened 2025-07-17
DM-51855: Enable color correction in Piff for ComCam.

PFLeget:


config/comCam/finalize_characterization.py
```diff
@@ -4,11 +4,11 @@
 # See DM-49673 for u-band.
 config.psf_determiner['piff'].spatialOrderPerBand = {
     "u": 2,
-    "g": 4,
-    "r": 4,
-    "i": 4,
-    "z": 4,
-    "y": 4,
```
arunkannawadi, line 11: Is lowering the order necessary purely for speed reasons, or do you run into overfitting issues?

arunkannawadi (approved):
I'd suggesting splitting the changes into two commits - one for reducing the order, one for toggling `useColor`.

config/comCam/finalize_characterization.py
PFLeget: It is for speed reasons and likely to avoid some overfitting. There are cross-terms between color and position that are not physical, but fitted anyway. There is a branch in Piff that should avoid that. I tested it for fitting ITL height but not yet for color. I will do that once this is merged.

PR #669, opened 2025-10-06
DM-52771: Correct filter name on 20250609

timj:


timj:
Ok. I will correct all four. Butler didn't care about them but I agree it would be confusing if someone looked at a header.

timj:
Pushed the fix.

PR #710, opened 2026-03-13
DM-51841: Add fiducial m5 for lsstCam

kadrlica:
Adding a fiducial m5 is necessary for calculating the effective time.


ehneilsen (approved):
Perfect, thank you!

PR #713, opened 2026-04-08
DM-54607: It does not look like LSSTCam has keywordsToCompare set in ISR

czwa:
Add cameraKeywordsToCompare for LSSTCam.


aferte (approved):
This looks good to me, thanks!

